PR #4, opened 2016-04-28
Create hello-world exercise

cswagner:
Adds the hello-world exercise to the Kotlin track.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@
 bin/configlet
 bin/configlet.exe
 CHECKLIST
+build
+.gradle
+.idea
+*.iml
--- a/config.json
+++ b/config.json
@@ -5,7 +5,7 @@
   "active": false,
   "test_pattern": "TODO",
   "problems": [
-
+    "hello-world"
   ],
   "deprecated": [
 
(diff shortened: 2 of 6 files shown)


kytrinyx:
I'm not sure it makes sense to stick the gradlew / gradlew.bat files in the exercise itself. I seem to recall going through some iterations of figuring this out in the Java track.  Since the CLI currently gets the exercise data in JSON format (each file is an entry in the JSON) I think it might have blown up with the scripts. Or not. I don't recall.

@exercism/java do any of you recall what happened with this?

jtigger:
Before I arrived on the scene, the decision had been made not to use the wrapper, but to walk folks through the Gradle install (or have them use IntelliJ's built-in support).  Very likely for the reason you state: we don't deliver JARs through the CLI.

In fact, what's good for the goose...  might want to take a look at the xjava build structure.

cswagner:
Thanks for the feedback. I've updated to not use the Gradle wrapper.

kytrinyx:
Thanks!

ErikSchierboom:
Great news! Looking forward to this track

PR #10, opened 2016-06-08
pangram: Add test case with repeated letters in both upper and lower cases

stkent:
My initial solution to the pangram problem was as follows:

``` kotlin
fun isPangram(sentence: String): Boolean {
    return sentence
            .filter { it.isLetter() }
            .toList()
            .distinct()
            .count() >= 26
}
```

This passes all the existing tests, but fails the newly-added test because I did not explicitly ignore case before calling `.distinct`.

### DIFF
--- a/exercises/pangram/src/test/kotlin/PangramTest.kt
+++ b/exercises/pangram/src/test/kotlin/PangramTest.kt
@@ -43,6 +43,11 @@ class PangramTest {
     fun mixedCaseAndPunctuation() {
         assertTrue(Pangrams.isPangram("\"Five quacking Zephyrs jolt my wax bed.\""))
     }
+    
+    @Test
+    fun mixedCaseDuplicatedCharacters() {
+        assertFalse(Pangrams.isPangram("the quick brown fox jumps over the lazy FOX"))
+    }
 
     @Test
     fun nonAsciiCharacters() {


sdavids13:
I believe it's legitimate to have the upper case also count as lowercase characters. Maybe your implementation would merely need:

```
fun isPangram(sentence: String): Boolean {
    return sentence.toLowerCase()
            .filter { it.isLetter() }
            .toList()
            .distinct()
            .count() >= 26
}
```

These tests were taken from the canonical test suite: https://github.com/exercism/x-common/blob/master/pangram.json so we would need to see if this should be pushed out to the rest of the languages. 

CC @kytrinyx

stkent:
Oh yes, my implementation was definitely at fault here, and your fix is the difference between passing and failing the new test. Adding the new test to the default suite covers off the possibility that others may fall into a similar trap in which an implementation _appears_ to be valid in the general case but isn't.

stkent:
Related: is there any way to know which language test suites were generated using the canonical suite, and which were written manually? Thanks!

sdavids13:
I believe the ruby track has their tests generated from files, I'm not sure about all of the others. It would be nice to build a test generator for the Kotlin track, but it does get a little tricky.

sdavids13:
OHHH, I misread the test! I thought you were just adding FOX to the end of the statement, but in fact you replaced `dog` with `FOX`. Maybe we should update the test name to say: `mixedCaseDuplicatedCharactersMissingDog`, or perhaps would we get the same point across by updated the test to: `assertTrue(Pangrams.isPangram("the quick brown fox jumps over the lazy dog FOX"))`

stkent:
On inspection I think that

```kotlin
assertTrue(Pangrams.isPangram("the quick brown fox jumps over the lazy dog FOX"))
```

would still pass with my original faulty implementation (I used inequality there as I was not sure how the German alphabet characters would be handled). The key is to disallow upper and lower case versions of the same character from being counted separately. I'm happy to rename to your suggestion (`mixedCaseDuplicatedCharactersMissingDog`) and push up again if that works for you?

kytrinyx:
I agree that if it's possible to submit an "obviously incorrect" solution, then our tests are not good enough, and that should be addressed in x-common (which I see that @sdavids13 is fixing already. Excellent).

behrtam:
Started a second try with https://github.com/exercism/x-common/pull/439 (after @sdavids13 failed) to get this added.

PR #26, opened 2017-01-15
Flatten RNA transcription into a top-level function.

jtigger:
Kotlin allows the programmer to define top-level functions.  `rna-transcription` is an excellent opportunity to showcase this language feature: there's exactly one function in this exercise.

### DIFF
--- a/exercises/rna-transcription/src/example/kotlin/RnaTranscription.kt
+++ b/exercises/rna-transcription/src/example/kotlin/RnaTranscription.kt
@@ -1,12 +1,9 @@
-object RnaTranscription {
-
-    fun ofDna(strand: String) = strand.map {
-        when (it) {
-            'A' -> 'U'
-            'G' -> 'C'
-            'C' -> 'G'
-            'T' -> 'A'
-            else -> it
-        }
-    }.joinToString("")
-}
+fun transcribeToRna(dna: String): String = dna.map { nucleotide ->
+    when (nucleotide) {
+        'A' -> 'U'
+        'G' -> 'C'
+        'C' -> 'G'
+        'T' -> 'A'
+        else -> nucleotide
+    }
+}.joinToString("")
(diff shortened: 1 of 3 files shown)


sdavids13:
I think it may be better to define the actual return type in the method definition:

```
fun transcribeToRna(dna: String): String = ""
```

```
fun transcribeToRna(dna: String): String = dna.map { nucleotide ->
    when (nucleotide) {
        'A' -> 'U'
        'G' -> 'C'
        'C' -> 'G'
        'T' -> 'A'
        else -> nucleotide
    }
}.joinToString("")
```

I am personally a bit leary about top-level functions but I guess this is a decent case for them. One thing we may want to do with the starter implementation as well is to provide an educational comment, perhaps along the lines of:
```
/**
 * This exercise demonstrates Kotlin's top-level functions
 * which doesn't require a class to provide a function.
 * These functions are available within the same package automatically,
 * they can be imported if outside of the implementing package.
 *
 * See: https://kotlinlang.org/docs/reference/functions.html#function-scope
 */
```

jtigger:
I'm hearing that anything global gives you pause as well; I'm with ya.  Since this is a stand-out feature of the language, it seems pretty important to highlight that option.

As you'll see in the PR, I tend to put hint-style comments in the test suite.  This way, there's no chance it gets submitted; making the appropriate thing the easiest thing.

jtigger:
Agreed on explicitly setting the return type, thank you!  Will update when I'm at my 'puter.

jtigger:
Hey @sdavids13.

In part of our dialogue, above, I pointed out that I deliberately put the comment in the test, not the implementation.  I also explained why I made that choice.

You did not respond to that comment.  Instead, you merged the PR and then moved that comment into the implementation class (in 7e09e4a0d4535976a028684b1621ef17d47e689b).

I find it frustrating that I put in the effort to not just provide the PR, but took the time to articulate my rationale for a design choice only to have the maintainer unilaterally undo that choice with no acknowledgement, explanation or better yet, discussion.  This is approach is not collaborative and discourages participation.

Please take the time to review [Maintaining an Exercism Language Track](https://github.com/exercism/exercism.io/blob/master/docs/maintaining-a-track.md); it expresses our values in terms of pragmatic advice.

sdavids13:
Oops, my apologies I missed that part of the comment. I will revert back the changes I made.

PR #27, opened 2017-01-16
etl: adjust tests to match description

jtigger:
The description for this exercise establishes a background where the program is transforming scrabble score configuration.  However, all but the last test in the suite use whole words as keys.  While *technically* correct, this difference between the description and the test suite represents accidental complexity rather than a challenge inherent in the exercise, itself.

The word-based implementation is a remnant of an implementation choice that was propagated but since rejected: exercism/xcpp#68.

This is the same work as on other tracks:

- exercism/xerlang#64
- exercism/xobjective-c#6


sdavids13:
Should the method definition be `transform(old: Map<Int, Collection<Char>>): Map<Char, Int>` instead of `transform(old: Map<Int, Collection<String>>): Map<String, Int>`?

jtigger:
It should!

PR #31, opened 2017-02-07
Rework hello-world as a directed introduction to the track.

jtigger:
- step-by-step tutorial gently introducing the practitioner to the language, the TDD process and Exercism.
- the tests now flow so that each test draws shape in the solution.
- moved `hello()` to the top-level, illustrating Kotlin's functions-as-first-class-citizens feature.

PR #32, opened 2017-02-07
Update installation and getting started instructions.

jtigger:
This complements #31, directing the user to the tutorial.

It also updates the instructions for installing the runtime; we've gotten far fewer setup questions when using package managers and these instructions are well-tested.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,154 +1,250 @@
-## Windows
+# Installing Kotlin
 
-If you are new-ish to programming in Java, these instructions are for you. This is a step-by-step opinionated guide to getting from zero to submitting your first exercise.
+In addition to the exercism CLI and your favorite text editor, practicing with Exercism exercises in Kotlin requires
+
+* the **Java Development Kit** (JDK) — Kotlin compiles to Java bytecodes; you need to install the JDK which includes both a Java Runtime *and* development tools (most notably, the Java compiler); and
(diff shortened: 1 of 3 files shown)


sdavids13:
Do you think it would be beneficial to show how to run the tests via IntelliJ in the tests doc?

jtigger:
I know what I'm about to say sounds like I'm out in left field and totally lost my marbles...

... but I went around the world on this one, last year — walking through the onboarding flow from the Exercism homepage up into `hello-world`; wrote the entire onboarding in terms of Intellij — and realized that (stay with me, here)  the best first experience for newcomers is the command line.

The key insight came when I tried to imagine what it would be like for someone brand new to the language and IDE.  Sure, it's absolutely intuitive for me _now_, but when I rolled back my years of experience on the tool... back to when I first laid eyes on an IDE, I vaguely remember feeling particularly overwhelmed.  There's actually quite a learning curce to the unitiated.

The programs our practitioners are writing in Exercism are tiny.  They actually won't see a whole lot of benefit from the power tool that is Intellij, at first.

Now, if they want to do any serious programming with Kotlin, they would _really_ want to jump into the IDE and climb that learning curve, no question.

Until then, the command-line tools (and the Gradle support is just great) are a far less steep and therefore more accessible.

See also: exercism/xjava#132

sdavids13:
Reasonable enough explanation for me.

PR #33, opened 2017-02-08
Provide setup instructions for contributors.

jtigger:
- perform a Journey Test as part of CI (and to support simplified local
  test of the same scope).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,16 @@ language: java
 jdk:
   - oraclejdk8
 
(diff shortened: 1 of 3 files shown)


sdavids13:
Much better than the TODO that has been there for a while now ;) 

@kytrinyx can you please add @jtigger as a committer on this track?

jtigger:
Thanks, @sdavids13!  I'm delighted to be an official part of this team!

PR #36, opened 2017-02-10
Fix gradle source sets so IDEs work for the non-generated tests

sdavids13:


jtigger:
Slick!  Do you work on Android apps, Steve?  This level of facility with Gradle feels like the Android Gradle plug-in level understanding.  :)

I was annoyed with having to run the build each time I wanted to run and editted test.

sdavids13:
Not an Android dev, but Gradle is my build tool of choice for all my Java apps. I have done a fair amount of custom build plugins and scripts for various scebarios.

PR #39, opened 2017-03-02
React test lambda destructuring

sdavids13:
The original test implementation for the react test via #29 was specifying multiple constructors in order to specify [different lambda parameter counts](https://github.com/exercism/xkotlin/pull/29#discussion_r100225524) which allowed for a test to be written like: `reactor.ComputeCell(one, two) { x, y -> x + y * 10 }`. The multiple constructor approach was ditched in favor of a [single constructor but taking in a list](https://github.com/exercism/xkotlin/pull/29#discussion_r100233867) into a lambda expression which resulted in a test of: `reactor.ComputeCell(one, two) { it[0] + it[1] * 10 }`

Now that Kotlin 1.1 has come out (#38) we can provide a similar test syntax originally provided by using the new lambda destructuring capability to provide a test of: `reactor.ComputeCell(one, two) { (x, y) -> x + y * 10 }`.

### DIFF
--- a/exercises/react/src/test/kotlin/ReactTest.kt
+++ b/exercises/react/src/test/kotlin/ReactTest.kt
@@ -36,7 +36,7 @@ class ReactTest {
         val reactor = Reactor<Int>()
         val one = reactor.InputCell(1)
         val two = reactor.InputCell(2)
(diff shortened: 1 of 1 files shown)


petertseng:
👍

PR #41, opened 2017-03-21
Add bank account exercise

nithia:
Loosely based on the C# and scala versions


sdavids13:
This looks great, I wonder if we should take the opportunity to showcase the new [coroutines capability](https://kotlinlang.org/docs/tutorials/coroutines-basic-jvm.html) in the tests instead of all of the threading. Here is a patch that achieves the same thing but looks marginally cleaner:

```
diff --git a/exercises/bank-account/build.gradle b/exercises/bank-account/build.gradle
index 532a683..a2f84db 100644
--- a/exercises/bank-account/build.gradle
+++ b/exercises/bank-account/build.gradle
@@ -10,8 +10,11 @@ buildscript {
 
 apply plugin: 'kotlin'
 
+kotlin.experimental.coroutines = 'enable'
+
 repositories {
     mavenCentral()
+    jcenter()
 }
 
 dependencies {
@@ -19,6 +22,7 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile "org.jetbrains.kotlin:kotlin-test-junit:$kotlin_version"
+    testCompile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:0.12'
 }
 test {
     testLogging {
diff --git a/exercises/bank-account/src/test/kotlin/BankAccountTest.kt b/exercises/bank-account/src/test/kotlin/BankAccountTest.kt
index 24282ee..af47c95 100644
--- a/exercises/bank-account/src/test/kotlin/BankAccountTest.kt
+++ b/exercises/bank-account/src/test/kotlin/BankAccountTest.kt
@@ -1,10 +1,9 @@
 import org.junit.Ignore
 import org.junit.Test
 import java.util.*
-import java.util.concurrent.CountDownLatch
-import kotlin.concurrent.thread
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
+import kotlinx.coroutines.experimental.*
 
 class BankAccountTest {
 
@@ -47,28 +46,25 @@ class BankAccountTest {
     @Ignore
     @Test
     fun concurrentBalanceAdjustments() {
-        val threads = 100
+        val threads = 1000
         val iterations = 500
-        val startLatch = CountDownLatch(1)
-        val endLatch = CountDownLatch(threads)
         val random = Random()
 
         val account = BankAccount()
 
-        repeat (threads) {
-            thread {
-                startLatch.await()
+        val jobs = List(threads) {
+            launch(CommonPool) {
                 repeat(iterations) {
                     account.adjustBalance(1)
-                    if (random.nextBoolean()) Thread.yield()
+                    delay(random.nextInt(10).toLong())
                     account.adjustBalance(-1)
                 }
-                endLatch.countDown()
             }
         }
 
-        startLatch.countDown()
-        endLatch.await()
+        runBlocking {
+            jobs.forEach { it.join() }
+        }
 
         assertEquals(0, account.balance)
     }
```

nithia:
Thanks, that's a good idea 👍

nithia:
@sdavids13 I have patched the test as you suggested and it does work rather nicely. Coroutines are a great addition to Kotlin :)

sdavids13:
Awesome, thanks for the contribution!

PR #43, opened 2017-04-17
Throw exception instead of returning "0000000000"

nithia:


exercises/phone-number/src/test/kotlin/PhoneNumberTest.kt
```diff
@@ -32,21 +33,19 @@ class PhoneNumberTest {
     }
 
     @Ignore
-    @Test
+    @Test(expected = IllegalArgumentException::class)
     fun invalidWhenOnly11Digits() {
-        val expectedNumber = "0000000000"
         val actualNumber = PhoneNumber("21234567890").number
```
sdavids13: We should just call the constructor without the `.number` it is unclear where the exception is being thrown (it should be done at object creation as it is currently).

exercises/phone-number/src/test/kotlin/PhoneNumberTest.kt
```diff
@@ -32,21 +33,19 @@ class PhoneNumberTest {
...
-        assertEquals(expectedNumber, actualNumber)
+        fail("IllegalArgumentException should have been thrown")
     }
 
     @Ignore
-    @Test
+    @Test(expected = IllegalArgumentException::class)
     fun invalidWhen9Digits() {
-        val expectedNumber = "0000000000"
         val actualNumber = PhoneNumber("123456789").number
```
sdavids13: Same comment w/r/t the `.number` here as well.

exercises/phone-number/src/test/kotlin/PhoneNumberTest.kt
nithia: Indeed, you are right. I will fix this

sdavids13:
Awesome, thanks.

PR #50, opened 2017-05-27
clock: add to track

stkent:
A port of the [xjava implementation](https://github.com/exercism/xjava/tree/master/exercises/clock/src).

[Canonical data](https://github.com/exercism/x-common/blob/54c3b7442709c9d269dbc434c072cf294cf9a7b8/exercises/clock/canonical-data.json)


exercises/clock/src/test/kotlin/ClockEqualTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+/*
+ * version: 1.0.1
+ */
+class ClockEqualTest {
+
+    @Ignore
+    @Test
+    fun clocksWithSameTimeAreEqual() {
+        assertTrue(Clock(15, 37) == Clock(15, 37))
```
sdavids13: Would it be more clear to just use `assertEquals` and `assertNotEquals` for these tests?

exercises/clock/src/example/kotlin/Clock.kt
```diff
@@ -0,0 +1,46 @@
...
+    companion object {
+        private val MINUTES_IN_AN_HOUR = 60
+        private val HOURS_IN_A_DAY = 24
+    }
+
+    init {
+        sanitiseTime()
+    }
+
+    override fun equals(other: Any?): Boolean {
```
sdavids13: If you make this a `data class` you don't need to override the equals operator, it is already taken care of for you as well as equals/hashcode/toString (though you will still want to override the default toString for this case).

exercises/clock/src/example/kotlin/Clock.kt
```diff
@@ -0,0 +1,46 @@
...
+
+        val minutesOverflow = minutes / MINUTES_IN_AN_HOUR
+        minutes %= MINUTES_IN_AN_HOUR
+        hours = (hours + minutesOverflow) % HOURS_IN_A_DAY
+    }
+
+}
+
+private fun Int.toTimeString(): String {
+    return toString().padStart(length = 2, padChar = '0')
```
sdavids13: 👍  That's pretty slick, I didn't know that was built-in.

exercises/clock/src/test/kotlin/ClockEqualTest.kt
stkent: Yeah, nice! I'll make the same change in xjava!

exercises/clock/src/example/kotlin/Clock.kt
stkent: 👍 willfix!

exercises/clock/src/example/kotlin/Clock.kt
stkent: Same, I searched somewhat optimistically and was pleasantly surprised :)

sdavids13:
Thanks for the work on this.

PR #51, opened 2017-05-28
Reorder track

### LINKED ISSUE #49: Proposal: update track order and exercise difficulty estimates to match xjava
Thinking:

0) Kotlin is about to receive a lot more attention now that it's a first-class language on Android; any high-leverage changes that improve the Exercism Kotlin experience are especially important at this moment in time;
1) Recent Java track order changes to incorporate estimated exercise difficulty have resulted in notably lower drop-off during the early exercises; the current order in xkotlin does not account for exercise difficulty;
2) The difficult of exercises is likely to be very similar between the two languages;
3) It's easy to refine order later on (this is an ongoing process in xjava).

stkent:
Closes #49.


config.json
```diff
@@ -4,7 +4,10 @@
   "repository": "https://github.com/exercism/xkotlin",
   "active": true,
   "deprecated": [
-
+    "accumulate",
+    "binary",
+    "hexadecimal",
+    "strain"
```
stkent: Updated deprecated exercises list to avoid contributors spending time on them.

config.json
```diff
@@ -24,248 +27,228 @@
...
+        "Reactive programming",
+        "Generics",
+        "Nested classes",
+        "Higher-order functions"
+      ]
     },
     {
       "difficulty": 1,
-      "slug": "pascals-triangle",
+      "slug": "perfect-numbers",
```
stkent: All subsequent exercises have not yet received difficulty estimates/placements in xjava, hence their positioning here.

config.json
```diff
@@ -4,7 +4,10 @@
   "repository": "https://github.com/exercism/xkotlin",
   "active": true,
   "deprecated": [
-
+    "accumulate",
```
sdavids13: Where does it say accumulate or strain is deprecated?

stkent: https://github.com/exercism/x-common/issues/553#issuecomment-279275406
https://github.com/exercism/x-common/issues/588#issuecomment-279275431

stkent: I can port over `list-ops` if you'd prefer not to have a gap between those exercises being deprecated and the replacement being implemented.

sdavids13: If you have time that would be great, otherwise I can try to get to it tomorrow.

stkent: I will have time Monday, so if you're bored tomorrow then go ahead, but if not I'll take care of it then. If you do start work please reply here so we don't double-dip (and I'll do the same :)

PR #55, opened 2017-05-29
collatz-conjecture: add to track

stkent:
New exercise! [x-common directory](https://github.com/exercism/x-common/tree/51f1341e4dfd070535e4d0e0039652a98cea7dc8/exercises/collatz-conjecture).


exercises/collatz-conjecture/src/example/kotlin/CollatzCalculator.kt
```diff
@@ -0,0 +1,12 @@
+object CollatzCalculator {
```
stkent: Not sure what the prevailing philosophy is here (within the track, or within the Kotlin community as a whole) for picking top-level function vs object-scoped function vs class-scoped function vs...!? Direction appreciated :)

nithia: I'm not sure if there's a general consensus on this, but most of the time I would pick an object-scope function unless the solution needs some state, in which case I'd make a class.

stkent: Thanks for the input!

PR #57, opened 2017-06-05
list-ops: add to track

stkent:
~~**Includes #64, do not merge before that**~~

Replaces the exercises deprecated in #64.

[`x-common` data](https://github.com/exercism/x-common/tree/da6b039ecf8dfc1a9bc8b29759c848cb0c94ab6e/exercises/list-ops).


exercises/list-ops/src/test/kotlin/ListExtensionsTest.kt
```diff
@@ -0,0 +1,141 @@
+import org.junit.Ignore
+import org.junit.Test
+import kotlin.test.assertEquals
+
+/*
+ * version: 1.0.0
```
stkent: `xjava` does not follow the canonical test suite (I suspect it may predate it) so I went ahead and wrote the Kotlin tests and example implementation fresh. I can port to `xjava` once we've ironed out any kinks!

exercises/list-ops/src/test/kotlin/ListExtensionsTest.kt
```diff
@@ -0,0 +1,141 @@
...
+                2.0,
+                emptyList<Double>().customFoldLeft(2.0) { acc, current -> current / acc })
+    }
+
+    @Ignore
+    @Test
+    fun testFoldLeftOnNonEmptyList() {
+        assertEquals(
+                64.0,
+                listOf(1.0, 2.0, 3.0, 4.0).customFoldLeft(24.0) { acc, current -> current / acc })
```
stkent: I was forced into using doubles for these fold left/right tests, since otherwise the integer division means we'll run into a division-by-0 error. I can add a tolerance to this assertion if you'd like (though it seems to work without 😕). I'm not sure if there's a way to keep this purely-int that I'm missing?

sdavids13:
This looks great to me, I like that you did this via extension functions. I would suggest that we include some hints to the users similar to what C-Sharp did here: https://github.com/exercism/xcsharp/blob/master/exercises/list-ops/HINTS.md but also include the link to Kotlin's extension function documentation: https://kotlinlang.org/docs/reference/extensions.html


Also, I think the difficulty should be bumped a little bit, maybe a 2 or 3?

exercises/list-ops/src/test/kotlin/ListExtensionsTest.kt
sdavids13: I feel like the division is super complicated, perhaps we should start out with an easier test that is a bit more straight-forward like fold left add, fold right add, fold left subtract, fold right subtract, etc such as: https://github.com/exercism/xpython/blob/master/exercises/list-ops/list_ops_test.py#L77

stkent:
@sdavids13 good points! I just pushed a new commit that includes exercise hints - let me know what you think of the content.

Regarding the difficulty rating; I'd be inclined to tweak that after #51 and #57 are both merged, so that it's easier to estimate how the difficulty of this exercise relates to all the others. My guess is that it will be closer to a 10 than to a 1, since extensions and generics combined make for some tricky spots.

stkent:
...which reminded me to tweak some of the tests so that they actually exercise the generic nature of the implementations :)

exercises/list-ops/src/test/kotlin/ListExtensionsTest.kt
stkent: Somehow I missed this comment first time around. I agree and will suggest that change upstream also!

stkent:
@sdavids13 upstream issue created; what would you like to do next? 🎁 

1. Hold this PR (and maybe #51?) until that upstream issue is resolved one way or the other?
2. Go ahead and modify this PR right now to include one/some/all of the tests mentioned in the upstream issue, on the basis that we believe this track should include them regardless of whether or not they exist in the canonical data?
3. Merge this PR right now, and update again later if the upstream issue resolution involves adding more tests?

sdavids13:
I think option 2 would be best.

stkent:
Roger that; I'll ping you when I've made those changes (could be a couple of days)!

exercises/list-ops/src/test/kotlin/ListExtensionsTest.kt
stkent: Tackling this shortly.

stkent:
@sdavids13 this is now updated with intermediate fold tests; however, this PR now depends on #64 (to reduce merge conflicts) so re-review should wait until that PR is merged and this PR is rebased.

sdavids13:
Both look good to me, merge away.

PR #60, opened 2017-06-10
Delete ignored key from config.json

kytrinyx:
Since the exercise implementations are all in the exercises directory
we no longer need to ignore any non-exercise directories in the root
of the track.

See https://github.com/exercism/meta/issues/3 for context.

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,14 +6,6 @@
   "deprecated": [
 
   ],
-  "ignored": [
-    "_template",
-    "bin",
-    ".gradle",
-    "docs",
-    "img",
-    "build"
-  ],
   "foregone": [
 
   ],


kytrinyx:
I've created another PR, #61, which normalizes all of the whitespace in the config.json so
that the change in this PR is minimized and easier to see.

If you merge the other PR first, I'll go ahead and rebase this one onto master.

Sorry about the noise!

stkent:
Looking into the failure now

stkent:
Hm, it looks similar to https://youtrack.jetbrains.com/issue/KT-15783 but that appears to have been resolved in Kotlin 1.1.0. I'm going to turn off [incremental compilation](https://kotlinlang.org/docs/reference/using-gradle.html#incremental-compilation) in master, which shouldn't be a performance hit because we compile fresh each time (in fact, it might even be a performance boost) and see if that resolves the issue.

config.json
```diff
@@ -6,14 +6,6 @@
   "deprecated": [
 
   ],
-  "ignored": [
-    "_template",
-    "bin",
-    ".gradle",
-    "docs",
-    "img",
-    "build"
```
stkent: I've noticed a `build` dir popping up in the exercises directory when running the unit tests, but if that remains an issue after merging this we can squash it separately.

kytrinyx: ah, interesting. Yeah, let's figure out what's going on with that when we next see it happen.

stkent: It's generated as part of the run-all-the-unit-tests script, but it's ignored so should never appear in this repo. I'll add a delete step to the unit tests script to avoid configlet getting upset on repeated runs of said script. (I think that only affects local development.)

kytrinyx: Nice, a delete step is an excellent idea

stkent: Implemented in #65 :)

stkent: Let's just never make an exercise with the slug `build` 😆

kytrinyx: LOL, let's not do that :)

PR #61, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.

### DIFF
--- a/config.json
+++ b/config.json
@@ -21,222 +21,310 @@
     {
       "difficulty": 1,
       "slug": "hello-world",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "gigasecond",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "rna-transcription",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "etl",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "word-count",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "accumulate",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "pangram",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "anagram",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "isogram",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "hamming",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "bob",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "perfect-numbers",
-      "topics": []
+      "topics": [
+
+      ]
     },
     {
       "difficulty": 1,
       "slug": "leap",
-      "topics": []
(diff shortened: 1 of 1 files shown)


stkent:
This broke in some interesting way. I'm guessing some `configlet` change, maybe, since when I run the unit tests script locally I also see failures with no changes in master. Whatever the issue is, it's the same in `xkotlin`, so I assigned myself to all the related issues while I investigate.

See https://github.com/exercism/configlet/compare/v1.1.0...v2.0.0 for the relevant configlet changes.

stkent:
Confirmed, if I update the `unit-tests` script to explicitly use configlet v1.1.0 (the prior release to today's v2.0.0) then unit tests run fine. Reading that diff more closely...

kytrinyx:
@stkent I made those changes in Configlet—how can I help?

kytrinyx:
> config.json does not include [_template].

Where is `_template`?

stkent:
@kytrinyx the output I'm seeing is this warning:

```
config.json does not include [_template].
```

However, I see it in the current config file:

```
"ignored": [
  "_template",
  ".gradle",
  "build",
  "docs",
  "gradle",
  "scripts"
],
```

I have just started to look at the configlet diff posted above; perhaps your more experienced eyes can spot any regression there faster than I will be able to? ❤️ 🤞

kytrinyx:
This is the problem: https://github.com/exercism/xkotlin/tree/master/exercises/_template

@stkent The assumption we make is that every directory within `exercises` is an exercise. If you place the `_template` in the root, that would fix this.

stkent:
There's a `_template` dir nested in `exercises` that is meant to be the seed for new exercises.

stkent:
Ahh, gotcha, thanks!. Which line of configlet is hiccuping there (for my own edification)?

kytrinyx:
It's not a configlet problem, it's a conceptual problem. We used to tell Configlet to ignore directories named certain things (per the config.json). When we removed that, configlet assumes that every single directory it finds within the exercises directory is an exercise. The code is here: https://github.com/exercism/configlet/blob/master/configlet/track.go#L100-L123

kytrinyx:
The change that caused this to fail is removing the IgnoredDirs() here: https://github.com/exercism/configlet/pull/17/files#diff-40e62472ea4139af33eebcc8f65668ebL94

stkent:
Ah ha! In that case I would humbly suggest that for clarity we might want to update the error message:

```
config.json does not include [_template].
```

to something more explicit like:

```
Exercise named _template found in exercises directory but not declared in the exercises block in config.json.
```

(This assumes I caught your drift properly (finally).)

kytrinyx:
> (This assumes I caught your drift properly (finally).)

Haha yes. You did.

So... you could also stick it in "deprecated" in config.json, which would mean that configlet would stop complaining about it.

I will update the error message, though. That's a good call.

stkent:
I've relocated the template to the root dir in both xkotlin and xjava; that seems cleaner than deprecating a fake exercise. Thanks for taking on the error message update!

stkent:
@kytrinyx if you could please rebase this against master, the CI build should now pass!

kytrinyx:
w00t! Thank you!

stkent:
Build is timing out; Travis seems 🐌 tonight. Trying again tomorrow.

stkent:
Build still timing out today 🤔. Gonna play with some speedups here. @sdavids13 can you see anything in the [recent commit history](https://github.com/exercism/xkotlin/commits/master) that might be slowing down recent builds?

stkent:
I'm going to rebase so it builds within the 50 minute limit!

kytrinyx:
> I'm going to rebase so it builds within the 50 minute limit!

A most excellent measure!

sdavids13:
Are there going to be merge conflict headaches with https://github.com/exercism/xkotlin/pull/51?

stkent:
Yes, but I'm happy to take care of fixing up #51.

PR #62, opened 2017-06-12
Sync scripts and travis config with xjava

stkent:
This results in significant speed improvements and better support for Windows. Commentary inline.


.gitignore
```diff
@@ -7,8 +7,3 @@ build
 .gradle
 .idea
 *.iml
-
-gradle/
-gradlew
-gradlew.bat
```
stkent: We're using these now! Updating to a modern version of Gradle (2.2.1 -> 3.5 wrapper) buys us a ~3x speed increase.

.travis.yml
```diff
@@ -3,16 +3,36 @@ language: java
 jdk:
   - oraclejdk8
 
+# http://docs.travis-ci.com/user/migrating-from-legacy
 sudo: false
-cache: bundler
```
stkent: Still cached, new syntax nearer the bottom of this file.

bin/build-jq.sh
```diff
@@ -0,0 +1,11 @@
...
+set -ex
+
+pushd bin
+curl --location https://github.com/stedolan/jq/releases/download/jq-1.5/jq-1.5.tar.gz >jq-1.5.tar.gz
+tar xvf jq-1.5.tar.gz
+cd jq-1.5
+./configure --disable-maintainer-mode && make
+mv jq ..
+popd
+
```
stkent: Building a newer version of jq from source to support Windows machines. See https://github.com/exercism/xjava/pull/277 for context.

bin/journey-test.sh
```diff
@@ -198,7 +216,7 @@ solve_all_exercises() {
 
   local track_root=$( pwd )
   local exercism_cli="./exercism --config ${exercism_configfile}"
-  local exercises=`cat config.json | jq '.exercises[] .slug' --raw-output`
+  local exercises=`cat config.json | jq '.exercises[].slug + " "' --join-output`
```
stkent: Windows compatibility.

bin/journey-test.sh
```diff
@@ -215,12 +233,12 @@ solve_all_exercises() {
 
     pushd ${exercism_exercises_dir}/${TRACK}/${exercise}
     # Check that tests compile before we strip @Ignore annotations
-    gradle compileTestJava
+    "$EXECPATH"/gradlew compileTestKotlin
     # Ensure we run all the tests (as delivered, all but the first is @Ignore'd)
     for testfile in `find . -name "*Test.${TRACK_SRC_EXT}"`; do
-      sed 's/@Ignore//' ${testfile} > "${tempfile}" && mv "${tempfile}" "${testfile}"
+      sed 's/@Ignore\(\(.*\)\)\{0,1\}//' ${testfile} > "${tempfile}" && mv "${tempfile}" "${testfile}"
```
stkent: Allows ignore comments. I'm guessing this disparity is why `@Ignore` annotations with descriptions were failing when I tried to use them in xkotlin before. Should handle both cases now.

bin/journey-test.sh
```diff
@@ -243,7 +261,7 @@ main() {
   local exercism_configfile=".journey-test.exercism.json"
   local xapi_port=9292
 
-  assert_installed "gradle"
```
stkent: Using the wrapper now.

bin/unit-tests.sh
```diff
@@ -10,6 +10,6 @@ bin/configlet .
 pushd exercises
 echo ""
 echo ">>> Running tests..."
-TERM=dumb gradle check compileStarterSourceKotlin --continue
+TERM=dumb ../gradlew check compileStarterSourceKotlin --parallel --continue
```
stkent: Parallelizes the unit tests; more easy speed!

stkent:
@sdavids13 since Travis seems to be running slow in general right now, I recommend reviewing this PR first (other PRs are failing due to timeouts after 50mins 😮). Even in slo-mode, Travis completed the builds of this branch in [less than 20 minutes](https://travis-ci.org/exercism/xkotlin/builds/241861125).

bin/journey-test.sh
sdavids13: Ah, that makes sense - I thought it was strange because I thought [this](https://github.com/exercism/xkotlin/blob/master/exercises/build.gradle#L51) would take care of it, but that is only if you run `gradlew test` on the root which then runs all the tests.

sdavids13:
This all looks good to me, much needed improvements.

bin/journey-test.sh
stkent: Ah, interesting, this is the first time I've seen that task. It's present in xjava too, albeit not updated to accommodate ignores with comments. I'll need to play around some to understand when that task is executed vs not, thanks for pointing it out!

stkent: Ok, so I think I understand this now:

- if we are running all tests from the 'unit' perspective, this copy-and-filter task is what creates a copy of the tests minus any ignore annotations in the `build/gen/test/java` directory, which is also specified as the location of the test source set;
- if we are running all tests from the 'journey' perspective, there's no 'root' project any more, so in order to simulate an implementor running all the tests, we need to manually strip ignore annotations in place.

Unfortunately the only way to avoid duplicating the 'strip ignore annotations' parts of these cases would appear to be packaging a copy task with the Gradle files that are served to implementors, which isn't great. As an alternative, maybe we should insert comments in both locations to help us remember that the two 'strip' implementations should be analogous. I'm thinking that's preferable.

sdavids13:
Yea, the comment should be sufficient for the time being.

stkent:
@sdavids13 comments added around those test invocations!

PR #63, opened 2017-06-13
Disable incremental compilation and unit test parallelization

stkent:
- Incremental compilation is not advantageous in a CI environment where
  builds start fresh each time;
- Parallelization of the unit tests was causing unexpected CI failures.
  I was unable to reproduce these failures locally. I tried several
  resolutions including running Kotlin compilation in-process, but none
  fully resolved the issue. Since CI stability is the priority and the
  speed increase obtained by parallelizing the unit tests was small, I
  chose to run the unit tests serially.

### DIFF
--- a/bin/unit-tests.sh
+++ b/bin/unit-tests.sh
@@ -10,6 +10,6 @@ bin/configlet .
 pushd exercises
 echo ""
 echo ">>> Running tests..."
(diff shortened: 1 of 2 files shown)


stkent:
Merging this on success so that we can get the other PRs building again!

PR #70, opened 2017-06-19
Remove invalid leading 1 from zip code

carpeliam:
The README specifies that the first number of the area code must fall between 2 and 9. This updates the examples to lead with a `2` instead of a `1` for the area code, and adds an extra test verifying that the area code does not begin with a `1`.


stkent:
Awesome, thanks for catching this and for opening up a PR! Looks good to me, I'll wait for @sdavids13 to confirm :)

PR #72, opened 2017-06-22
complex-numbers: add to track

stkent:
[Upstream directory](https://github.com/exercism/problem-specifications/tree/117d062a18124f7535124a50ff9bb7c6660aa4e7/exercises/complex-numbers).


exercises/complex-numbers/src/example/kotlin/ComplexNumber.kt
```diff
@@ -0,0 +1,26 @@
+data class ComplexNumber(val real: Double = 0.0, val imag: Double = 0.0)
+
+operator fun ComplexNumber.plus(other: ComplexNumber)
```
sdavids13: Would it make sense to put these functions inside the class instead of making them extension functions?

stkent: Does that work for operation functions? If so, yes! If not, I liked them outside so that the order of definition matched the order in which these methods were exercised by the tests. I'll have a play locally.

stkent: Updated!

PR #73, opened 2017-06-22
prime-factors: add to track

stkent:
[problem specification](https://github.com/exercism/problem-specifications/tree/117d062a18124f7535124a50ff9bb7c6660aa4e7/exercises/prime-factors).


config.json
```diff
@@ -199,6 +199,13 @@
     },
     {
       "difficulty": 5,
+      "slug": "prime-factors",
```
stkent: From the Java track.

exercises/prime-factors/src/example/kotlin/PrimeFactorCalculator.kt
```diff
@@ -0,0 +1,38 @@
...
+            while (remainder.rem(divisor) == 0L) {
+                result.add(divisor)
+                remainder /= divisor
+            }
+
+            divisor++
+        }
+
+        return result
+    }
```
stkent: I tried to avoid this duplication, but since `Number` isn't itself `Comparable`, I couldn't find a way to unify these definitions. Open to input here!

sdavids13: You could call the `primeFactors` method with an int cast to a long then map the `list<Long>` to an int via `list.map { it.toInt() }`

stkent: Good call, updated!

sdavids13:
What's with the `exercises/prime-factors/src/starterSource/kotlin/.keep` file? Feel free to merge these in if I take too long in response (sorry, have been a bit busy lately)

stkent:
The keep file keeps the sourceset's Kotlin directory from disappearing (https://stackoverflow.com/a/17929518/2911458). I stole from the Java track; if that's not something that's being done here, I will remove. (I use IntelliJ's auto directory generation, so it doesn't bother me!)

sdavids13:
Yea, I understand the purpose of the .keep files just wondering what files are supposed to go in that directory?

stkent:
Hmm, not sure; looking back at the configurations now it seems that's only used for the unit test script, and that starter source as delivered to the user should live in the main source set directory, right? In which case, yeah, I'll 💣 this keep file.

sdavids13:
Yea, I don't think that directory should be getting used for anything I thought, it should be safe to nuke the entire directory.

PR #77, opened 2017-06-24
secret-handshake: add to track

stkent:
[problem specification directory](https://github.com/exercism/problem-specifications/tree/b3daec17adfedcc7122328e2cb3a0987a6e34bb5/exercises/secret-handshake)


exercises/secret-handshake/src/example/kotlin/Signal.kt
```diff
@@ -0,0 +1,5 @@
+enum class Signal {
```
sdavids13: I don't believe this duplicate enum is necessary, I believe the `src/main/kotlin` classes are on the classpath.

exercises/secret-handshake/src/example/kotlin/HandshakeCalculator.kt
```diff
@@ -0,0 +1,23 @@
+object HandshakeCalculator {
+
+    private val REVERSE_SIGNALS_BIT_POSITION = 4
+
+    fun calculateHandshake(number: Int): List<Signal> {
+        val result = mutableListOf<Signal>()
+
+        Signal.values().filterTo(result) {
```
sdavids13: You can avoid the mutableList by just doing:

```
val result = Signal.values().filter { signal -> isBitSet(signal.ordinal, number) }

if (isBitSet(REVERSE_SIGNALS_BIT_POSITION, number)) {
    return result.asReversed()
}
return result
```

exercises/secret-handshake/src/example/kotlin/HandshakeCalculator.kt
```diff
@@ -0,0 +1,23 @@
+object HandshakeCalculator {
+
+    private val REVERSE_SIGNALS_BIT_POSITION = 4
```
sdavids13: For constants you generally put in a `companion object`

exercises/secret-handshake/src/example/kotlin/Signal.kt
stkent: Ah, then I can probably remove similar classes from several other exercises (and will do so!)

exercises/secret-handshake/src/example/kotlin/HandshakeCalculator.kt
stkent: This appears to be disallowed inside an `object`.

exercises/secret-handshake/src/example/kotlin/Signal.kt
stkent: Looks like this caused the unit test compilation to fail. Maybe I unwittingly changed something while consolidating with the Java track. Looking...

stkent: Hmm, that still runs `compileStarterSourceKotlin` as I would expect...

stkent: I guess I should ask: what are the various sourcesets used for? This is one of the areas in which the Kotlin and Java tracks diverge.

sdavids13: Oh, grr, it looks like I just did a soft link before for this same thing:

https://github.com/exercism/kotlin/blob/02ecc5a70719f0b0e741992f33879f880f7c407f/exercises/allergies/src/example/kotlin/Allergen.kt

stkent: Ah, yeah, Java was doing the same thing but that's not Windows-friendly. Happy to implement either option, let me know which you'd prefer!

sdavids13: I prefer to avoid the duplication, but not really tied to either. This PR looks good either way.

PR #84, opened 2017-07-10
rotational-cipher: add to track

stkent:
[Problem specification directory](https://github.com/exercism/problem-specifications/tree/73ebd13032219888ee29374b648b9a769d536549/exercises/rotational-cipher).

As part of this changed I moved `atbash-cipher` to earlier in the track. In the description for `rotational-cipher`, it makes reference to `atbash-cipher`, so this seemed to make sense. It also seems odd that `atbash-cipher` was previously ranked harder than e.g. `flatten-array`, `bracket-push`, etc.; some of those exercises feel like they are currently underscored.


exercises/rotational-cipher/src/example/kotlin/RotationalCipher.kt
```diff
@@ -0,0 +1,18 @@
+class RotationalCipher(val shiftKey: Int) {
+
+    companion object {
+        private val LOWERS = "abcdefghijklmnopqrstuvwxyz"
+        private val UPPERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
stkent: Alternatively:

```kotlin
private val LOWERS = ('a'..'z').joinToString("")
private val UPPERS = ('A'..'Z').joinToString("")
```

sdavids13: Or how about:
```
private val LOWERS = ('a' .. 'z').toList()
private val UPPERS = ('A' .. 'Z').toList()
```

I am cool with any of these options the rest of this pull request looks good, merge away whichever way you want to go.

PR #88, opened 2017-07-12
forth: add to track

stkent:
[problem specification directory](https://github.com/exercism/problem-specifications/tree/8bae822d7cc641be3037f78c74a0d78a72aced00/exercises/forth)


exercises/forth/src/example/kotlin/ForthEvaluator.kt
```diff
@@ -0,0 +1,139 @@
...
+
+                Pair("dup") { values ->
+                    require(values.isNotEmpty()) { "Duplicating requires that the stack contain at least 1 value" }
+                    values.push(values.peek())
+                },
+
+                Pair("drop") { values ->
+                    require(values.isNotEmpty()) { "Dropping requires that the stack contain at least 1 value" }
+                    values.pop()
+                    Unit
```
stkent: This feels _super_ weird. I needed something here because `values.pop` returns non-Unit. Is there a better way to do this?

stkent: I guess I could have the method signature use `Any` instead of `Unit`?

stkent:
Needs update w/ README also.

PR #100, opened 2017-08-11
Added further detail to space age exercise README

DarkPaladinUHC:
After I completed this exersize (without the rounding) I thought it would be nice to add this little notice so nobody else runs into the problem.

### DIFF
--- a/exercises/space-age/README.md
+++ b/exercises/space-age/README.md
@@ -12,7 +12,8 @@ Given an age in seconds, calculate how old someone would be on:
    - Neptune: orbital period 164.79132 Earth years
 
 So if you were told someone were 1,000,000,000 seconds old, you should
-be able to say that they're 31 Earth-years old.
+be able to say that they're 31 Earth-years old. All returned values should be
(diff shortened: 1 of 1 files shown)


stkent:
This one slipped by me somehow! Thanks for the PR :)

I like the change. That README file is actually generated based on a bunch of other data, so the next time we regenerate it we're at risk of losing this change if we merge this pull request as-is. The best way to avoid that would be to make this same change in the [shared problem specifications repository](https://github.com/exercism/problem-specifications/). I'll open a discussion issue there tomorrow to see if this is something we'd like all tracks to incorporate. Once a consensus is reached in that `problem-specifications` issues, we'll know how to proceed here (merge locally, or simply regenerate, based on whether the same change is applied in the shared repository or not).

Sorry for the wall of words; let me know if anything is unclear!

DarkPaladinUHC:
Sounds good

stkent:
I've regenerated the READMEs based on upstream changes, so I'm going to go ahead and close this PR. Thanks for bringing this improvement to our attention!

PR #107, opened 2017-09-02
Double indent code blocks to allow numbered lists to not break

Stargator:
[Markdown syntax](https://daringfireball.net/projects/markdown/syntax#list) states code blocks should be double indented in order for the numbering in the lists to continue.

Seeing as how the Kotlin Installation and Tests page had numbered lists that were broken. I opted to fix them. [The issue is discussed here for v2 of exercism](https://github.com/exercism/v2-feedback/issues/105). I am checking the v1 pages to see what are affected by this.

As you can see the current rendering of the Installation page has the numbering thrown off. This should fix that.

kotlin-tests

As an example, this is how it will render when formatted as done in this PR (minus no CSS):

kotlin-example

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -24,25 +24,25 @@ Open an administrative command prompt.  (If you need assistance opening an admin
 
 1. If you have not installed Chocolatey, do so now:
 
-       ```batchfile
-       C:\Windows\system32> @powershell -NoProfile -ExecutionPolicy unrestricted -Command "iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))" && SET PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin
-       ```
+        ```batchfile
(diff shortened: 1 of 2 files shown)


Stargator:
@sdavids13 I can fix that

sdavids13:
Thanks for the modifications, for some reason the github markdown reader doesn't seem to recognize some of these [code blocks](https://github.com/Stargator/kotlin/blob/149cd7f78395942fe2e67736713fffdee3b231b6/docs/INSTALLATION.md#windows). Happen to have any ideas? Here is a screenshot: 
screen shot 2017-09-04 at 11 54 01 pm

Stargator:
@sdavids13 Hmm... do you see the same thing for the pages I updated for Java and Dart tracks?

sdavids13:
No, doesn't look like it though it appears as though the indention on the dart and java track is 5 spaces instead of 8. This is making me think that it needs single indention instead of double?

Stargator:
@sdavids13 Not going to GitHub [who were the ones I had originally contacted](https://github.com/Stargator/dart/pull/1) and the [Markdown syntax guide I linked above](https://daringfireball.net/projects/markdown/syntax#list)

Stargator:
@sdavids13 Made a few changes, should address that. It does seem like it doesn't quite line up with the syntax, but it works....

Stargator:
@sdavids13 Rebased the branch to address a conflict

Stargator:
@sdavids13 Does this work for you?

sdavids13:
Yea, looks good.

PR #110, opened 2017-10-10
two-fer: implementation of two-fer exercise in Kotlin

Nik-Mouz:


Nik-Mouz:
In Gitter it was recommended to me by @Nobbz to also change the hello-world exercise to be updated to the new style of hello-world, (which simply asks the user to return the string "hello world!"), instead of the current implementation that requires string manipulation. This then allows string manipulation to be introduced to the user via the two-fer exercise that is implemented in this PR.

stkent:
Hi! Thanks for this contribution!

I'll add review comments inline in a second, but I also wanted to post some information relevant to the `hello-world` update you are tackling. Here are the PRs where we updated `hello-world` in the Java track: https://github.com/exercism/java/pull/663 and https://github.com/exercism/java/pull/666. They should give you some pointers as to what updates may need to happen to _supporting_ files in the Kotlin track. Feel free to reach out if you have any questions!

exercises/two-fer/.meta/src/reference/kotlin/Twofer.kt
```diff
@@ -0,0 +1,10 @@
+fun twofer(name: String = ""): String {
```
stkent: Since we don't need to allow `null` input here, I'd say the closest interpretation of "no name given" would be invoking the `twofer` function with no parameter, a case that is currently not exercised by the tests. To implement that, we should probably update this signature to

```kotlin
fun twofer(name: String = "you"): String
```

and have this exercise focus on a more typical use of default parameter values. Then the first test should mimic https://github.com/exercism/java/blob/master/exercises/two-fer/src/test/java/TwoferTest.java#L16-L22 but with no parameter passed instead of a `null` parameter passed.

The Java track debated what to do with an empty string input, and settled on being very literal about handling it: https://github.com/exercism/java/pull/733. I'd say the same boundary case could be exercised here, since invocation with no parameters is the "right way" to call this function without specifying a name. Any explicit input can therefore be interpreted as something the caller wants to display as the name (even if empty). Does that make sense to fresh eyes?

exercises/two-fer/.meta/src/reference/kotlin/Twofer.kt
```diff
@@ -0,0 +1,10 @@
...
+    var tempName = name
+    val regex = Regex("[A-Za-z]")
+    val matched = regex.containsMatchIn(tempName)
+
+    if(!matched)
+        tempName = "you"
+
+    return "One for $tempName, one for me."
```
stkent: In light of the above, this sample implementation can be reduced to a single line.

exercises/two-fer/src/test/kotlin/TwoferTest.kt
```diff
@@ -0,0 +1,24 @@
+import org.junit.Ignore
+import org.junit.Test
+import org.junit.Assert.assertEquals
```
stkent: Just a note that https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-equals.html is probably available - I assume it's a wrapper around the JUnit assertion anyway though!

Nik-Mouz:
Hey @stkent, I believe my latest commit has completed your requested changes? If there is something I am missing or more changes required please let me know. Thanks! :+1:

stkent:
Looking at this tomorrow!

exercises/two-fer/.meta/src/reference/kotlin/Twofer.kt
```diff
@@ -0,0 +1,3 @@
+fun twofer(name: String = "you"): String {
+    return "One for ${if (name.isBlank()) "you" else name}, one for me."
```
stkent: I think this can be altered to

```kotlin
return "One for $name, one for me."
```

since supplying a blank string as input implies different intent to providing _no_ input.

exercises/two-fer/src/test/kotlin/TwoferTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    fun noNameGiven() {
+        assertEquals("One for you, one for me.", twofer())
+    }
+
+    @Test
+    @Ignore
+    fun emptyStringGiven() {
+        assertEquals("One for you, one for me.", twofer(""))
+    }
```
stkent: Let's mimic the Java tests and (1) have the expected string here be "One for , one for me.", and (2) move this test to the end of the suite.

exercises/two-fer/src/test/kotlin/TwoferTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Ignore
+    fun anotherNameGiven() {
+        assertEquals("One for Bob, one for me.", twofer("Bob"))
+    }
+
+    @Test
+    @Ignore
+    fun aNameGiven() {
+        assertEquals("One for Alice, one for me.", twofer("Alice"))
+    }
```
stkent: I believe this test should be moved above `anotherNameGiven`?

PR #112, opened 2017-10-18
Adds Signal type to empty ArrayList in tests

gypsydave5:
I was just doing this exercise when I noticed that it failed to compile unless a `Signal` type annotation was added to the `emptyList()`

This PR adds that annotation. Hope it helps.


gypsydave5:
Actually, I was being a bit of a fool - works just fine if I return a List, not an ArrayList. Sorry for any trouble! 😄

stkent:
I plan to (eventually) update the Kotlin track to be more like the Java one and provide stubbed interfaces for tricky problems such as this one. Reverse engineering signatures from tests is not a common part of software development! Thanks for calling this out; it reminds me of the need to improve in this area 👍

PR #235, opened 2018-07-07
test added: multiples of 3 or 5 up to 20

lotharschulz:
adding a test case using the data from readme file. 15 is a multiple of 5 and 3 and smaller than 20. However, 15 counts only once when summing up to 78.

### DIFF
--- a/exercises/sum-of-multiples/src/test/kotlin/SumOfMultiplesTest.kt
+++ b/exercises/sum-of-multiples/src/test/kotlin/SumOfMultiplesTest.kt
@@ -11,6 +11,12 @@ class SumOfMultiplesTest {
 
     @Test
     @Ignore
+    fun `multiples of 3 or 5 up to 20`() {
+        assertEquals(78, SumOfMultiples.sum(setOf(3, 5), 20))
+    }
+
+    @Test
+    @Ignore
     fun `multiples of 3 or 5 up to 4`() {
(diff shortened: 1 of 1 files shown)


lotharschulz:
@vmichalak: please provide a status update, when this change can be merged

lotharschulz:
Thanks for coming back.

I proposed this PR because I believe the existing test case does not test an interesting case:

The result can be **78** or **93** given a set of 2 items `setOf(3, 5)` and the natural number of _20_ as threshold. The reason for this is: 15 is part of the interim results for both, `3` and `5`.
I believe 15 should be counted only once for the final result, so I am a fan of the final result **78**. In case you'd count 15 twice, you'd end up with a final result of **93**

This debug screenshot hopefully helps to understand my point of view:

screenshot_2018-10-02_08-33-36


All other existing test cases passed in my tests with both implementations. The additional test case I propose shall make sure only one implementation as described above is valid and thus shape the expected overall result.

I am happy to provide a PR to https://github.com/exercism/problem-specifications/blob/master/exercises/sum-of-multiples/canonical-data.json if you agree in general with my approach.

sjwarner-bp:
Right, I understand better now @lotharschulz - thanks for such a great update!

I think that normally each language track tries to adhere to the problem specification. What I would suggest is raising an issue/PR over in that repo where the maintainers for that can take your suggestion into consideration. That way, if this change was accepted, it would get applied to all of our language tracks!

Does that make sense? 🙂

PR #237, opened 2018-07-30
Change invalid test requirements

pindab0ter:
Can't calculate products with a zero width span or input.


sjwarner-bp:
Hi @pindab0ter - sorry, but just to make sure, is this PR updating the `series` exercise to be in line with the [most recent canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/series/canonical-data.json) for the exercise? It looks like it might be, but I wanted to double check before going further :slightly_smiling_face:

PR #240, opened 2018-08-07
Add rail-fence-cipher exercice

vmichalak:
It's my first contribution to Exercism: I ported the rail-fence-cipher exercice.
If there is a problem, please let me know


exercises/rail-fence-cipher/.meta/version.txt
```diff
@@ -0,0 +1 @@
+1.2.40
```
sjwarner-bp: Is this correct? 40 seems awfully high :laughing: 

I think the file should also just be called `version` without the `.txt`

vmichalak: Oh damn, it's not the Kotlin version here ? :3

vmichalak:
No problem, I will continue to wear exercises and when the construction will work again, we will merge it :)

vmichalak:
No problem, I will continue to port exercises and when the build will work again, we will merge it :)

PR #241, opened 2018-08-08
Add reverse-string exercise.

vmichalak:


exercises/reverse-string/.meta/src/reference/kotlin/ReverseString.kt
```diff
@@ -0,0 +1 @@
+fun reverse(input: String): String = input.reversed()
```
sjwarner-bp: Some maintainers dislike the use of built-in functions, though I personally have little problem with it seeing as learning library functions is an important part of learning a language IMO! 🙂

config.json
```diff
@@ -16,6 +16,16 @@
       ],
       "auto_approve": true
     },
+    {
+      "slug": "reverse-string",
+      "uuid": "0000e75e-9b43-11e8-98d0-529269fb1459",
+      "core": false,
+      "unlocked_by": null,
```
sjwarner-bp: Is this deliberately a 'bonus exercise' or is it worth specifying an `unlocked_by`?

exercises/reverse-string/.meta/src/reference/kotlin/ReverseString.kt
vmichalak: I changed the implementation in a new commit.

config.json
vmichalak: I think it's better to make it a bonus exercise no ?

vmichalak:
Wow, Sorry i forget this PR. It's now ok normally :)

mikegehard:
Thank you!

I squashed the commits down into one here: https://github.com/exercism/kotlin/commit/7b3c307ba4e4c049db58cf40a32c71603110d282

vmichalak:
Thanks @mikegehard

PR #247, opened 2018-09-04
Unlock leap by gigasecond

katrinleinweber:
Hello,

as a Kotlin newbie (with mostly R & Python experience; no Java) I stumbled a lot over the very early demand for advanced concepts (like [custom accessor of a property](https://exercism.io/my/solutions/74ecab0faaae46598406791bbde81ece) in [`leap`](https://github.com/exercism/kotlin/commit/4a2979ba0ae2150c3ffe86200139dacb857feac4)).

Since [`gigasecond`](https://github.com/exercism/kotlin/commit/b93e1f5ff71bf30634812285624ff88f653d7a76) is the first core exercise whose tests prescribe that concept, how about letting it unlock `leap`? I haven't looked at the other unlocks by `hello-world`, but the same idea may apply there.

### DIFF
--- a/config.json
+++ b/config.json
@@ -31,7 +31,7 @@
       "slug": "leap",
       "uuid": "639e59b6-84aa-4f13-9718-537606703c43",
       "core": false,
-      "unlocked_by": "hello-world",
+      "unlocked_by": "gigasecond",
       "difficulty": 2,
       "topics": [
         "control_flow_if_else_statements",


katrinleinweber:
In case a formality like "unlocked exercise shouldn't be easier" needs to be satisfied: yes. However, I feel that an alternative commit, which broadens the test case(s) would be more useful overall.

I suggest to avoid pushing learners into advanced concepts where simpler ones are valid solutions. At least that is what friend with a decade of Java experience plus a year of Kotlin wondered about, when I showed them the test for this exercise.

sjwarner-bp:
There isn't a formal rule there, I was just curious.

As for test suites, we try to mirror the canonical-data in exercism/problem-specifications. Do you think the tests aren't quite right, or the actual implementation is over-complicated? :slightly_smiling_face:

PR #248, opened 2018-10-01
update ci scripts

sjwarner-bp:
CI here is broken.

Hopefully I can fix it up! 🙂


sjwarner-bp:
CI now passes!

The breakage was caused by a change to configlet tooling in the new version of exercism.

I have changed the CI script to be a more traditional test runner. This should work, though I'll keep an eye on it!

PR #263, opened 2018-10-05
added version update and new tests

luoken:
updated the tests and the version.


luoken:
I updated the test but didnt update the correct version. it should be up to date now if you would like to double check.

PR #264, opened 2018-10-05
Added exercise run-length-encoding based on Java track

lathspell:
Another challenge from the Java track.


config.json
```diff
@@ -431,6 +431,19 @@
         "matrices"
       ]
     },
+    {
+      "slug": "run-length-encoding",
+      "uuid": "4499a3f9-73a7-48bf-8753-d5b6abf588c9",
```
sjwarner-bp: Hi @lathspell - is this a new UUID or has this been taken from Java?

exercises/run-length-encoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+```text
+"AABCCCDEEEE"  ->  "2AB3CD4E"  ->  "AABCCCDEEEE"
+```
+
+For simplicity, you can assume that the unencoded string will only contain
+the letters A through Z (either lower or upper case) and whitespace. This way
+data to be encoded will never contain any numbers and numbers inside data to
+be decoded always represent the count for the following character.
+
+# Running the tests
```
sjwarner-bp: Please generate the README using configlet, to avoid the extra Java content here.

exercises/run-length-encoding/src/test/kotlin/RunLengthEncodingTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    fun setUp() {
+        runLengthEncoding = RunLengthEncoding()
+    }
+
+    @Test
+    fun encodeEmpty() {
+        assertEquals("", runLengthEncoding.encode(""))
+    }
+
+    @Test
```
sjwarner-bp: Please `@Ignore` uninitial tests

mikegehard:
Hello @lathspell. I am wondering if you have some time to get back to take a look at these?

If not, I'm going to close this PR as I'm trying to clean up the repo since it has gone a bit stale.

Issue #170, opened 2018-06-08
series: update test suite and version file to 1.0.0


sjwarner-bp:
Please add a version file `exercises/series/.meta/version` with the string `1.0.0` for the series exercise.

Since implementing this exercise, [canonical-data](https://github.com/exercism/problem-specifications/commits/master/exercises/series/canonical-data.json) has been added for `series` in the problem-specifications repo. In order ensure our tests cover all cases and are consistent with other tracks, we should try to mirror problem-specifications as closely as possible.

Because of this, [the test suite](https://github.com/exercism/kotlin/blob/master/exercises/series/src/test/kotlin/SeriesTest.kt) should be rewritten to comply with version 1.0.0.

Additionally, the [reference implementation](https://github.com/exercism/kotlin/blob/master/exercises/series/.meta/src/reference/kotlin/Series.kt) will need updating to pass the new test suite.

sophiekoonin:
I'll pick this one up

PR #273, opened 2018-10-16
Add version for Series and update tests and implementation to match c…

sophiekoonin:
Fixes issue #170. I've used `return when` syntax - hope that's ok for Kotlin beginners!


mikegehard:
Thank you!

PR #276, opened 2018-10-22
Fix gradle-school exercise test

tehbilly:
The `expected` value is incorrect, causing the test to fail. This is also true with the reference implementation of the exercise.


mikegehard:
Thank you!

PR #278, opened 2018-10-24
two-fer: updated readme

uzilan:
Updated two-fer readme according to https://github.com/exercism/problem-specifications/pull/1374


mikegehard:
Thank you @uzilan!

Issue #185, opened 2018-06-08
complex-numbers: update tests and version file (v1.0.0 -> v1.3.0)


sjwarner-bp:
Please update the version specified in the complex-numbers version file from 1.0.0 to 1.3.0.

This update has been caused by the application of a new input policy, and the addition of a new test case.

Please additionally update the test suite to match the [canonical-data](https://github.com/exercism/problem-specifications/commits/master/exercises/complex-numbers/canonical-data.json). The test described by "Exponential of a number with real and imaginary part" should be added, and the reference solution updated if necessary.

DanCorder:
I can take a look at this

PR #279, opened 2018-12-17
Update complex numbers test to match canonical data - issue #185

DanCorder:
The first change addresses issue #185 updating to v1.3.0 of the canonical data.

The second change is not strictly needed, but it seems nicer to use Kotlin's standard libraries.


mikegehard:
Thank you!

PR #281, opened 2019-01-31
Add test to confirm no duplicates

mikegehard:
The README talks about this but we never write a test for it. I screwed up an implementation and this test caught the duplicates.


mikegehard:
Commit here: https://github.com/exercism/kotlin/commit/9868d7843bfebf73c2c7080cf45e909a62dba3c5

PR #285, opened 2019-03-03
Clarify what functions you are not supposed to use.

### LINKED ISSUE #280: list-ops: it's not clear what existing functions/operators you can use
The current instructions for the list-ops say:

> In functional languages list operations like length, map, and reduce are very common. Implement a series of basic list operations, without using existing functions.

AFAIK, Kotlin does not provide a way to do anything with a list without using functions.  Even built in operators like `[]` wind up calling functions (e.g. `get`).  So I don't get how it's possible to do this exercise without using at least some of the built in Kotlin or Java functions.

I've looked over [most of the community solutions](https://exercism.io/tracks/kotlin/exercises/list-ops/solutions) and all the ones I've looked at use some existing functions like `drop`, `+`, `isEmpty`, etc.

Can the instructions get clarified to be more specific about what functions you are allowed to use?

myronmarston:
Closes #280.

### DIFF
--- a/exercises/list-ops/README.md
+++ b/exercises/list-ops/README.md
@@ -4,7 +4,7 @@ Implement basic list operations.
 
 In functional languages list operations like `length`, `map`, and
 `reduce` are very common. Implement a series of basic list operations,
-without using existing functions.
+without using the existing functions of the same name.
 
 ## Hints
 


mikegehard:
Thank you!

PR #292, opened 2019-03-13
Update robot-name RobotTest.kt

matthewmcgarvey:
I was able to make it through this challenge using a computed property like:

```kotlin
val name : String
    get() = createName()
```

which doesn't make sense


mikegehard:
Thanks for the PR!

Merged here: https://github.com/exercism/kotlin/commit/2e37e5028643eac3ac00f4b3d50e54c1a29df037

PR #294, opened 2019-04-06
Fix for test on Empty Stirng

antimoroma:
For empty string we must have as expected result

One for you, one for me


NobbZ:
To be honest, I consider the empty string as a valid name, since kotlin has other means to differ between no value given and a value given.

The current implementation might be a conscious choice of the original implementor of the exercise.

Issue #297, opened 2019-04-12
Checklist: normalise topics


SleeplessByte:
**Note**: the suggestions are generated automagically. If something is _actually_ missing, please open an issue at [problem-specifications](https://github.com/exercism/problem-specifications) to add the missing topics.

```
=> fetching config from ../exercism-kotlin/config.json
=> reading topics from ..\problem-specifications\TOPICS.txt
```

<details>
  <summary>List of valid topics</summary>

  - `abstract_classes`
  - `callbacks`
  - `conditionals`
  - `events`
  - `equality`
  - `exception_handling`
  - `generics`
  - `inheritance`
  - `interfaces`
  - `loops`
  - `operator_overloading`
  - `pattern_matching`
  - `polymorphism`
  - `recursion`
  - `threading`
  - `type_conversion`
  - `variables`
  - `arrays`
  - `booleans`
  - `classes`
  - `dates`
  - `discriminated_unions`
  - `enumerations`
  - `floating_point_numbers`
  - `graphs`
  - `integers`
  - `lists`
  - `maps`
  - `matrices`
  - `optional_values`
  - `queues`
  - `records`
  - `sequences`
  - `sets`
  - `stacks`
  - `strings`
  - `structs`
  - `time`
  - `trees`
  - `tuples`
  - `algorithms`
  - `bitwise_operations`
  - `concurrency`
  - `cryptography`
  - `domain_specific_languages`
  - `files`
  - `filtering`
  - `games`
  - `globalization`
  - `logic`
  - `math`
  - `memory_management`
  - `networking`
  - `parallelism`
  - `parsing`
  - `pattern_recognition`
  - `performance`
  - `randomness`
  - `refactoring`
  - `regular_expressions`
  - `searching`
  - `security`
  - `sorting`
  - `text_formatting`
  - `transforming`
  - `behavior_driven_development`
  - `functional_programming`
  - `object_oriented_programming`
  - `reactive_programming`
  - `test_driven_development`
</details>

## Exercises with non-standard topics

- [x] [`leap`](https://github.com/exercism/kotlin/tree/master/exercises/leap)
  Incorrect topics:
  - [ ] `control_flow_if_else_statements` (perhaps: `conditionals`)
- [ ] [`sublist`](https://github.com/exercism/kotlin/tree/master/exercises/sublist)
  Incorrect topics:
  - [ ] `enumeration` (perhaps: `enumerations`)
[...]

PR #314, opened 2019-10-03
Replace non-standard topics

stkent:
Per #297 - normalizes topics from that issue (`sublist` topics have apparently been normalized since that issue was opened).

### DIFF
--- a/config.json
+++ b/config.json
@@ -43,7 +43,7 @@
       "unlocked_by": "gigasecond",
       "difficulty": 2,
       "topics": [
-        "control_flow_if_else_statements",
+        "conditionals",
         "integers"
       ]
     },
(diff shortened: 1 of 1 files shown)


stkent:
Should pass after #323 is merged.

dector:
@stkent Easy fix for: `sublist` (`enumerations`) can be added as well. :)

config.json
```diff
@@ -823,8 +823,8 @@
       "difficulty": 10,
       "topics": [
         "generics",
-        "higher_order_functions",
-        "nested_classes",
+        "functional_programming",
+        "classes",
```
dector: @eparovyshnaya what do you think about this. Should we use standard topics here?
I'm kinda ok with it.

eparovyshnaya: That's all about fitting to common metadata - it's mandatory. The other question is if this metadata rich enough to express significant pecularities. 

Both these cases are ok for me. `higher_order_functions` practically equals to `functional_programming` support, and `classes` covers `nested classes` with an appropriate context.

dector:
Triggering Travis re-check.

PR #320, opened 2019-10-08
issues 310: Create stub files for all exercises

eparovyshnaya:
- stub file for "Two Fer" task solution is appended
 - tests are un-ignored, as they got compilable over the main source base

---
The stub is questionable if the following aspects.

It contains hints to study about 
 - [internal scope](https://kotlinlang.org/docs/reference/visibility-modifiers.html), which is one of the label Kt features
- [single-expression function](https://kotlinlang.org/docs/reference/functions.html#single-expression-functions) which is concise and idomatic tool

It does not contain a clue for [default arguments](https://kotlinlang.org/docs/reference/functions.html#default-arguments). I just do not know how to mention it withount giving a precise solution.

It is quite a Kotlin-like, but I'm not sure if it fits Exercism studying methodics. 

Do you wish the stub to stay as close to the reference solution as possibl? 
Do you want to avoid distraction by extra study motivators (like *internal* keyword and *=* sign in a function signture)? 
If you do, I will rework the stub for sure. 
If you don't, I have anoher proposal: can I add these links in the function's kotlin-doc?

### DIFF
--- a/exercises/secret-handshake/src/main/kotlin/HandshakeCalculator.kt
+++ b/exercises/secret-handshake/src/main/kotlin/HandshakeCalculator.kt
@@ -1,0 +1,5 @@
+object HandshakeCalculator {
+    fun calculateHandshake(number: Int): List<Signal> {
+        TODO("Implement the function to complete the task")
+    }
+}
--- a/exercises/two-fer/src/main/kotlin/Twofer.kt
+++ b/exercises/two-fer/src/main/kotlin/Twofer.kt
@@ -1,0 +1,3 @@
+internal fun twofer(name: String): String {
+    TODO("Implement the function to complete the task")
+}


SleeplessByte:
Why are the tests un-ignored? Only the first test should be enabled.

IMO, the stub does _not_ need to compile, but I think it's find to give a stub _without_ a default argument. It's a pretty natural path to add a default argument as you go.

eparovyshnaya:
All questionable things are undone. 

We stll cannot get this compiled until [PR 318](https://github.com/exercism/kotlin/pull/318)  is merged.

exercises/secret-handshake/src/main/kotlin/HandshakeCalculator.kt
```diff
@@ -0,0 +1,5 @@
+object HandshakeCalculator {
+    fun calculateHandshake(number: Int): List<Signal> {
+        return emptyList()
+    }
+}
```
SleeplessByte: You might want to consider adding an empty newline at the EOF, I think this is common for most editors to insert, and will not pollute git changes when the final line is changed.

exercises/two-fer/src/main/kotlin/Twofer.kt
```diff
@@ -0,0 +1,3 @@
+internal fun twofer(name: String): String {
+    return ""
+}
```
SleeplessByte: Same as the other file.

exercises/two-fer/src/main/kotlin/Twofer.kt
```diff
@@ -0,0 +1,3 @@
+internal fun twofer(name: String): String {
+    return ""
```
SleeplessByte: Instead, I suggest throwing a [not implemented error](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-not-implemented-error/index.html).

exercises/secret-handshake/src/main/kotlin/HandshakeCalculator.kt
```diff
@@ -0,0 +1,5 @@
+object HandshakeCalculator {
+    fun calculateHandshake(number: Int): List<Signal> {
+        return emptyList()
```
SleeplessByte: Instead, I suggest throwing a [not implemented error](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-not-implemented-error/index.html).

SleeplessByte:
> If you don't, I have anoher proposal: can I add these links in the function's kotlin-doc?

If you show me what that would look like... sure :)

exercises/secret-handshake/src/main/kotlin/HandshakeCalculator.kt
eparovyshnaya: Have sense. There is also a specially dedicated [TODO](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-t-o-d-o.html) shortcut. Does not it fir more?

SleeplessByte:
Nice!

Don't use KDoc for this, but the [TrackInsert hint](https://github.com/exercism/configlet#hints)! Then, readme needs to be [re-generated](https://github.com/exercism/configlet#generate). The hint will become _part of `README.md`_.

Yes, use `TODO`!! That's a great fit.

eparovyshnaya:
Got it. Will remove the KDocs from both stubs to call it a day for this PR. Will open another one for TrackHints.

SleeplessByte:
@dector if you squash-merge here, those steps are unnecessary.

dector:
@SleeplessByte Kinda. I'm not using built-in functionality here intentionally.
I'm not sure how it handles commit author info (especially, for signed commits). There is some [open issues](https://github.com/isaacs/github/issues/1368). I'm just not familiar how and what this merge mode does behind the curtain. :)

There is no rush for this PR so I suggested to do it manually.

eparovyshnaya:
Guys, I do appreciate your help and care. Just let me sort it out. Taming this *rebase* became an affair of honor for me.

eparovyshnaya:
@dector, made it. Can you please merge?

PR #323, opened 2019-10-16
updated travis jdk to openjdk11

uzilan:


mikegehard:
What do you think about updating the [installation docs](https://exercism.io/tracks/kotlin/installation) to talk about Java 11? That way we are testing with the version of the JDK that we have folks install?

uzilan:
Yeah I updated the documentation in the Java repo as well, see https://github.com/exercism/java/pull/1768
I'm can use those docs in here too, if you think it's ok.

Issue #288, opened 2019-03-10
Upgrade to the latest Kotlin 


mikegehard:

blueglyph:
I'm not sure why this was open, but I'll use it to comment, sorry if that was not meant for that.

Usually I'm changing the build.gradle setup file like this before using it, to work around the fixed, old Kotlin version. Perhaps this is how it could be set in all exercises?

```
buildscript {
  ext.kotlin_version = '1.+'
  repositories {
    mavenCentral()
  }
  dependencies {
    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
  }
} 

```

(instead of specifying the Kotlin version explicitly)

The same can be used for the junit version.

mikegehard:
Thanks for the thoughts @blueglyph. Yeah we need to figure out a way to make the version a bit more flexible as the versions change.

dector:
I guess it's very possible to generate exercises based on some meta-data. :)

dector:
Proof of concept: https://github.com/dector/genercism/blob/master/data/exercises/reverse-string.json

dector:
Folks from Slack pointed me to [problem-specifications](https://github.com/exercism/problem-specifications). Will check it out and probably use it as a source data for generator.

dector:
Generated exercises snapshots are [here](https://github.com/dector/genercism/issues/1). Please, notice that exercises that are using Maps as arguments/return types are not compiling for now.

sshine:
@mikegehard, @uzilan: Just an FYI: If you bump Kotlin version constraints in `build.gradle` you may experience that this change doesn't propagate to the website; see e.g. exercism/haskell#864 for how we deal with it.

@dector wrote:
> I guess it's very possible to generate exercises based on some meta-data. :)

It's nice to see that you're getting acquainted with the Exercism infrastructure and try to expand upon it.

Perhaps the subject of auto-generating exercises based on problem-specifications deserves its own issue, rather than conflate this with bumping the Kotlin version for all existing exercises. (The task of this issue is more tangible and deserves few dependencies.) There are quite a few tracks that have good test generators; C#, F# and OCaml are among them; unfortunately, as it's not something I've dealt with directly, I cannot say what their state of documentation is. (OCaml track's test generator framework had a total makeover recently, but the changes still lack a bit of documentation.)

dector:
@sshine 

> There are quite a few tracks that have good test generators

[...]

PR #326, opened 2019-10-19
For #288. Update tooling for `reverse-string`.

dector:
1. Update Kotlin version to `1.3.50` (latest stable).
1. Update Gradle to `5.6.2`.
1. Use fail-fast mode for tests. In combination with test order this will improve testing process for those who are doing exercises (no need to edit test-file and remove `@Ignore` annotation).
1. Added missing source/build files for `reverse-string` exercise.


dector:
Depends on #323

dector:
Will close it for now and reopen as a template update proposal.

PR #339, opened 2019-11-12
#310: Create stub files for all exercises

eparovyshnaya:
stubs for b-tasks: 
 - bank-account
 - beer-song
 - binary
 - binary-search
 - bob

### DIFF
--- a/exercises/beer-song/src/main/kotlin/BeerSong.kt
+++ b/exercises/beer-song/src/main/kotlin/BeerSong.kt
@@ -1,0 +1,5 @@
+object BeerSong {
+    fun verses(startBottles: Int, takeDown: Int){
+        TODO("Implement the function to complete the task")
+    }
+}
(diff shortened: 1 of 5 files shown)


exercises/beer-song/src/main/kotlin/BeerSong.kt
```diff
@@ -0,0 +1,5 @@
+object BeerSong {
```
dector: We definitely should reduce amount of used objects in Kotlin track. :)

eparovyshnaya: Agree! Singleton is not the best pattern in the world. 
 
The signature is dictated by the current tests implementation and reference implementation. We'd need to rework them. Out of the scope of this task, but we have almost 30 exercised tested and ref-implemented as singletones (out of 68 in the repo).  Sigletone definitely does not worth such an attention. Is now a proper time to post an issue?

_If your test requires smth like `Pianist.play(4) == "Walzing Matilda"`, there is not much a student can do .
But one extraordinary fellow came from C-world and was anaware of singletones, so, as the first solution, he named a package with an upper-cased name (like `Pianist`), put a kt-file there, and implemented public-scoped function (like `play(Int)`) in it. And so the test was ashamed :)_

PR #342, opened 2019-11-13
#310: Create stub files for all exercises - change, clock. collatz-conjecture

eparovyshnaya:
- stubs for c-tasks: 
       - change 
       - clock
       - collatz-conjecture
  - redundant *.keep*-files are cleaned up

### DIFF
--- a/exercises/change/src/main/kotlin/ChangeCalculator.kt
+++ b/exercises/change/src/main/kotlin/ChangeCalculator.kt
@@ -1,0 +1,7 @@
+class ChangeCalculator {
+    // TODO: implement proper constructor
+
+    fun computeMostEfficientChange(grandTotal: Int): List<Int> {
+        TODO("Implement the function to complete the task")
+    }
+}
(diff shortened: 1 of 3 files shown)


exercises/change/src/main/kotlin/ChangeCalculator.kt
```diff
@@ -0,0 +1,7 @@
+class ChangeCalculator {
```
uzilan: Shouldn't the class contain an argument like `list: List<Int>`? It wont compile otherwise

eparovyshnaya: It contains `implement proper constructor` TODO-instruction instead. 
It's ok for a stub to not fit the signature requirements dictated by tests. It can be even uncompilable. And there are cases, when a correct signature is actually the solution.

uzilan: Ok cool

uzilan:
LGFM :)

eparovyshnaya:
Thanks for review, @uzilan !

PR #343, opened 2019-11-13
#310: Create stub files for all exercises - difference-of-square…

eparovyshnaya:


uzilan:
LGFM :)

exercises/difference-of-squares/src/main/kotlin/Squares.kt
```diff
@@ -0,0 +1,15 @@
+class Squares {
```
dector: Do we really need class here?

eparovyshnaya: I'm kinda guided my the appropriate test suite. 

We might be willing to shape a general convension and to rework all the tests suits/ref impls/stubs bundles according to it. But this all is out of the scope of this PR and the issue is is dedicated to. I can post a proper issue for further review/rework, but freeze here at what existing test suit requires. Do you agree?

dector: Awesome! You are totally right.

eparovyshnaya: @dector, issue #349

PR #345, opened 2019-11-14
For #310. Create stubs for exercises.

dector:
Some stubs created from referenced solutions. As discussed, sanitizing is out of the scope of this issue.


dector:
Each commit can be reviewed/merged separately.

Adding exercises that required some attention because of file renaming.

exercises/grade-school/src/main/kotlin/GradeSchool.kt
```diff
@@ -0,0 +1,17 @@
+class School {
```
eparovyshnaya: Why not naming the kt-file after the class? Isn't it a recommended Kotlin practice?
Or each solution file should be named after the slug?

exercises/roman-numerals/.meta/src/reference/kotlin/RomanNumerals.kt
```diff
@@ -1,4 +1,5 @@
-object RomanNumeral {
+object RomanNumerals {
+
```
eparovyshnaya: Colleagues, do we have common project code style settings files for idea/eclipse?

exercises/grade-school/src/main/kotlin/GradeSchool.kt
dector: Technically, `School` represents some entity here.

But I left all entity-names to avoid fixing tests as well.

exercises/roman-numerals/.meta/src/reference/kotlin/RomanNumerals.kt
dector: I guess we should use [official code style](https://kotlinlang.org/docs/reference/code-style-migration-guide.html#in-gradle).

eparovyshnaya: This one does not insert a new line between object/class declaration  and the first method, for example. Not a big deal, actually, when it comes to stubs.

dector: Ah, you mean that. In this case it's my personal taste - I think it look better when logical blocks (like function declaration) are not "stuck" to class header/declaration.

But I can see both styles in Kotlin std lib. :)

eparovyshnaya: If this empty line does not appear automatically as a result of code formatting - I'm perfectly ok with it. 
But if it is - I'd like to have the same formatter for understandable reasons.

PR #347, opened 2019-11-14
#310: Create stub files for all exercises -  kt-filenames

eparovyshnaya:
Rename kt-files containing stub implementation according to task name.


eparovyshnaya:
No functional fixes, only file renames

eparovyshnaya:
my fault, fixing

PR #354, opened 2019-11-18
bump `triangle` implementation to spec of 1.2.1 version

eparovyshnaya:
- Tests are alligned with the canonical data of version 1.2.1 (current). 
 - tests are renamed with backtics
 - test names are mosty formatted  like `suite name | case name`

> Triangle inequality violation

As it's described in *comments* section for the whole data set, in the case of triangle inequality violation each track can opt whether it's better to throw an exception or function normally. 
Further canonical test data describes specification for *function normally case*. 

Since Kotlin implements exception-drivel flow, *violation* tests are reworked to a sngle ctor invocation.


dector:
I like it!

PR #356, opened 2019-11-19
upgrade `saddle points` implementation to spec of 1.5.0 version

eparovyshnaya:
test suite is refactored in order to avoid boilerplated code

### DIFF
--- a/exercises/saddle-points/src/test/kotlin/MatrixTest.kt
+++ b/exercises/saddle-points/src/test/kotlin/MatrixTest.kt
@@ -6,136 +6,128 @@ import kotlin.test.assertEquals
 class MatrixTest {
 
     @Test
-    fun `single saddle point`() {
-        val matrix = Matrix(
-            listOf(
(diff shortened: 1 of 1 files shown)


exercises/saddle-points/src/test/kotlin/MatrixTest.kt
```diff
@@ -6,136 +6,128 @@ import kotlin.test.assertEquals
...
-        val matrix = Matrix(listOf(emptyList()))
-
-        val expectedSaddlePoints = emptySet<MatrixCoordinate>()
-
-        assertEquals(expectedSaddlePoints, matrix.saddlePoints)
-    }
+    fun `no saddle points for empty matrix`() =
+        assertSaddlePointsEqual(
+            Matrix(listOf(emptyList())),
+            emptySet<MatrixCoordinate>()
```
dector: nit: May be just `emptySet()` I guess.

PR #375, opened 2019-12-01
upgrade `atbash-cipher` test suite  to spec of 1.2.0 version

eparovyshnaya:


uzilan:
LGFM :)

PR #377, opened 2019-12-01
upgrade `isogram` test suite  to spec of 1.7.0 version

eparovyshnaya:


uzilan:
LGFM :)

PR #378, opened 2019-12-03
upgrade `phone-number` test suite and ref implementation to spec 1.7.0

eparovyshnaya:
- test set is altered and replenished according to the specification
 - tests are renamed with backticked notation
 - reference implementation is fixed to satisfy new specification requirement: *throw an error for invalid cases instead of returning `null`*


exercises/phone-number/.meta/src/reference/kotlin/PhoneNumber.kt
```diff
@@ -8,14 +8,15 @@ data class PhoneNumber(private val rawNumber: String) {
...
 
     init {
-        val containsInvalidChar = invalidChars.containsMatchIn(rawNumber)
-
-        if (containsInvalidChar) {
-            number = null
-        } else {
-            val digits = rawNumber.replace(Regex("[^\\d]"), "")
-            number = digitsValidationRegex.matchEntire(digits)?.groupValues?.last()
+        if (invalidChars.containsMatchIn(rawNumber)) {
```
uzilan: Maybe `require()` would be more fitting here?

eparovyshnaya: Indeed!

PR #379, opened 2019-12-04
upgrade `allergies` test suite to spec of 2.0.0 version

eparovyshnaya:
- test set is altered and replenished according to the specification


uzilan:
LGFM! I like the custom asserts :)
Well done!

PR #380, opened 2019-12-05
upgrade `pascals-triangle` test suite to spec of 1.5.0 version

eparovyshnaya:
- test set is diminished according to the specification: exceptional case processing is not required anymore, so here we also stop demanding it from a student


exercises/pascals-triangle/src/test/kotlin/PascalsTriangleTest.kt
```diff
@@ -1,123 +1,88 @@
...
-
-    @Rule
-    @JvmField
-    var expectedException: ExpectedException = ExpectedException.none()
-
     @Test
-    fun zeroRows() {
-        val expectedOutput = emptyList<List<Int>>()
+    fun `zero rows`() = assertTriangleEquals(
+            0,
```
dector: nit: Hmm. Seems like IDEA using 8 symbols for continuation indentation. I don't remember if I changed it manually to 4 (to comply detekt/ktlint rules) or just imported official style guide.

PR #382, opened 2019-12-06
upgrade `matching-brackets` test suite to spec of 2.0.0 version

eparovyshnaya:
- tests set is replenished according to teh specification


uzilan:
LGFM

PR #383, opened 2019-12-06
upgrade `isbn-verifier` test suite to spec of 2.7.0 version

eparovyshnaya:
- tests set is replenished according to teh specification
  - test names are backticked


exercises/isbn-verifier/src/test/kotlin/IsbnVerifierTest.kt
```diff
@@ -4,100 +4,74 @@ import kotlin.test.assertFalse
...
-        assertTrue(isbnVerifier.isValid("3-598-21507-X"))
-    }
+    fun `invalid | check digit is character other than X`() = assertInvalid("3-598-21507-A")
 
     @Ignore
     @Test
-    fun checkDigitIsACharacterOtherThanX() {
-        assertFalse(isbnVerifier.isValid("3-598-21507-A"))
-    }
+    fun `invalid | forbidden character`() = assertInvalid("3-598-P1581-X")
```
uzilan: Should we stick to the description in the canonical data? It is "invalid character in isbn"

eparovyshnaya: `invalid` is already used for the *case* prefix, so I just failed to find proepr synonym. Do you have a suggestion?

uzilan: Aha I see. No I guess it's fine then

PR #386, opened 2019-12-07
Added new exercise 'armstrong-numbers'

lathspell:
This solution is uglier than the Python one as in Java Char.toInt() gives the ASCII number not the digit value therefore Character.numbericValue() (or `c-'0'`) has to be used. Also Kotlin lacks a `Int.pow()` method (KT-8248) so we have Math.pow() and BigInteger.pow(). The former insists on using Doubles and I didn't like using possibly inaccurate floating point just for this.


eparovyshnaya:
@lathspell it looks like unkock-condition is incorrect 
>-> The exercise 'armstrong-numbers' is being unlocked by a non-core exercise. Non-core exercises can only be unlocked by core exercises.

PR #387, opened 2019-12-08
upgrade `clock` exercise test set to the specification of 2.4.0 version

eparovyshnaya:
- test set is replenished and restructured according to the spec


uzilan:
LGFM!

eparovyshnaya:
Thanks for review!

PR #388, opened 2019-12-08
Binary search tree

lathspell:
Added exercise based on Java and Python implementations.

Would be cleaner if the tree had to be initialized so that the root element could never be NULL but I wasn't sure if that would be conforming to canonical-data.json.

I'm unsure about the difficulty and order. Feel free to merge and then change it as you please! :)


exercises/binary-search-tree/README.md
```diff
@@ -0,0 +1,85 @@
...
+> Use `gradlew.bat` if you're on Windows
+
+In the test suites all tests but the first have been skipped.
+
+Once you get a test passing, you can enable the next one by removing the
+`@Ignore` annotation.
+
+## Source
+
+Josh Cheek [https://twitter.com/josh\_cheek](https://twitter.com/josh_cheek)
```
eparovyshnaya: Can you please pick up more appropriate and reliable link here?

lathspell: That's the line that was in https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search-tree/metadata.yml since 2016 and is present in the C#, Go, Python and Java READMEs as well :-)

But I understand your point. Maybe you ask the original spec author and we change it in all languages once the spec version is incremented?

eparovyshnaya: yes, reasonable idea

PR #392, opened 2019-12-10
Added new exercise 'custom-set'

lathspell:
The Java track gave this a stunning difficulty of 10 and omitted the empty class. Python and CSharp had the empty class and rated it 3-5. I went for difficulty 3 and an empty class that already contains the constructors and the "operator" annotations. This makes the implementation rather easy. Too easy? Comments?


exercises/custom-set/.meta/src/reference/kotlin/CustomSet.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun isSubset(other: CustomSet): Boolean = other.set.containsAll(set)
+
+    fun isDisjoint(other: CustomSet): Boolean = set.intersect(other.set).isEmpty()
+
+    fun contains(other: Int): Boolean = set.contains(other)
+
+    fun intersection(other: CustomSet): CustomSet = CustomSet(set.intersect(other.set))
+
+    fun add(other: Int) {
+        set = set + other
```
eparovyshnaya: Here is a tradeoff: 
  - define `set` as a `var` and reassign it here fully, 
  - or let it be *final* `val` inited as  `mutableSetOf()` and use `set.add()` in this method.

The second approach is always prefferable, as the class becomes *less mutable* - still not thread safe but at least represents always the sampe single entity in memory (set in our case).

exercises/custom-set/src/main/kotlin/CustomSet.kt
```diff
@@ -0,0 +1,40 @@
+class CustomSet(private var set: Set<Int> = setOf()) {
```
eparovyshnaya: Test set never requires constructing our class from a set. It's ok to use this two-constructor approach in ref implementation, but we should not demand it from a student, as it is not a part of spec. 

It looks like only `vararg`-based constructor is strictly required. And I'd better hide this clue to avoid a student taking it for granted. `Vararg` is definitely a concept a student must pay attention to.

Let's stub it lightly like 
```
class CustomSet {
    // TODO: implement proper constructor
    ...
}
```

exercises/custom-set/.meta/src/reference/kotlin/CustomSet.kt
lathspell: I played with both but couldn't decide :) I will change it to mutableSet and strip the constructor from the empty class as proposed!

PR #393, opened 2019-12-11
upgrade `robot-simulator` exercise implementation to specificati…

eparovyshnaya:
- test set is restructured according to the spec
 - stub and ref implementation are revised: it appears that the partial movement methods (`turnLeft`, `turnRight`, `advance`) are not required by the specification at all and, it their core, are just details of ref implementation. So we cannot force a student to stick with the same approach. Test set is reimplemented accordingly.


exercises/robot-simulator/src/test/kotlin/RobotTest.kt
```diff
@@ -5,194 +5,183 @@ import kotlin.test.assertEquals
...
-
-        robot.simulate("RRAAAAALA")
-
-        assertEquals(GridPosition(x = -3, y = -8), robot.gridPosition)
-        assertEquals(Orientation.SOUTH, robot.orientation)
-    }
+    fun `series of instructions - moving east and north`() =
+        Robot(GridPosition(x = 8, y = 4), Orientation.SOUTH)
+            .instructed("LAAARRRALLLL")
+            .should {
```
dector: :heart_eyes:

PR #395, opened 2019-12-11
Added new exercise 'dnd-character'

lathspell:


exercises/dnd-character/src/test/kotlin/DndCharacterTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    fun `random character is valid`() {
+        val char = DndCharacter()
+        assertTrue(char.strength >= 3 && char.strength <= 18)
+        assertTrue(char.dexterity >= 3 && char.dexterity <= 18)
+        assertTrue(char.constitution >= 3 && char.constitution <= 18)
+        assertTrue(char.intelligence >= 3 && char.intelligence <= 18)
+        assertTrue(char.wisdom >= 3 && char.wisdom <= 18)
+        assertTrue(char.charisma >= 3 && char.charisma <= 18)
```
eparovyshnaya: There are *ranges* in Kotlin, you can write more readable assert:
```
assertTrue(char.carisma in (3 .. 18))
```

and `with` scope-funtion, so you can make it even cleaner
```
fun `random character is valid`() = 
        with(DndCharacter()) {
            assertTrue(strength in (3 .. 18))
            assertTrue(dexterity in (3 .. 18))
            ....
        }
```

exercises/dnd-character/src/main/kotlin/DndCharacter.kt
```diff
@@ -0,0 +1,22 @@
+class DndCharacter {
+
+    val strength: Int = -1
+    val dexterity: Int = -1
+    val constitution: Int = -1
+    val intelligence: Int = -1
+    val wisdom: Int = -1
+    val charisma: Int = -1
+    val hitpoints: Int = -1
```
eparovyshnaya: `val smth = -1` means *forever*, that can confuse a student.
 Add *TODO*-notes, please :)

exercises/dnd-character/.meta/src/reference/kotlin/DndCharacter.kt
```diff
@@ -0,0 +1,21 @@
...
+    val wisdom: Int = ability()
+    val charisma: Int = ability()
+    val hitpoints: Int = 10 + modifier(constitution)
+
+    companion object {
+
+        fun ability(): Int = (1 .. 4).map { Random.nextInt(1,6) }.sorted().drop(1).sum()
+
+        // Alternative: java.lang.Math.floorDiv(score - 10, 2)
+        fun modifier(score: Int): Int = floor((score - 10.0) / 2).toInt()
```
eparovyshnaya: Both are ok, choose one and eliminate the comment line.

exercises/dnd-character/src/test/kotlin/DndCharacterTest.kt
lathspell: Indeed, much cleaner!

PR #396, opened 2019-12-15
upgrade `say` exercise implementation to specification of 1.2.0 version

eparovyshnaya:
 - for exceptional cases error message is not checked anymore


uzilan:
LGFM

PR #397, opened 2019-12-15
Added new exercise 'dominoes'

lathspell:
This one is a translation from the Java exercise. I changed the use of Collections.frequency() to Kotlin List.count{} to get rid of the hashCode/equals override functions in the Domino data class.


exercises/dominoes/src/test/kotlin/DominoesTest.kt
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Test
+    fun `can reverse dominoes`() {
+        val dominoesList = listOf(Domino(1, 2), Domino(1, 3), Domino(2, 3))
+        val chain = Dominoes.formChain(dominoesList)
+        assertValidChain(dominoesList, chain)
+    }
+
+    @Test
```
eparovyshnaya: Please, Use
```kotlin
@Test (expected=ChainNotFoundException::class)
```
and avoid `var expectedException`. We do not force a student to use a precise error message. The fact that exception is thrown is sufficient. 

I'd also recommend to have a look at updated tests, for example at *robot-simulator/RobotTest*. I can give you an idea of pleasure of local dsl and single-assert tests. Maybe you'll find it interesting for further test set development. It's not mandatory, though.

lathspell: Nice practice for me. Please have a look at the updated test.

PR #398, opened 2019-12-18
zebra-puzzle

lathspell:
This one was actually one of the hardest puzzles I've seen so far. I adapted the solution from the Java track as I was to lazy to actually solve it myself :) I did implement the Permutation algorithm though to get rid of the `java.util.stream.*` dependencies.


exercises/zebra-puzzle/src/main/kotlin/ZebraPuzzle.kt
```diff
@@ -0,0 +1,8 @@
+class ZebraPuzzle() {
+
+    var waterDrinker: String = TODO("Implement this class to complete the task")
+        private set
+
+    var zebraOwner: String = TODO("Implement this class to complete the task")
+        private set
```
eparovyshnaya: Here we share reference implementation details. But a student is not obliged to stick with our implementational approach. 

Can we declare two extra functions like
```kotlin
fun drinksWater(): String
fun ownsZebra(): String
```
and retarget test to  them? 

They will be return `waterDrinker` and `zebraOwner` respectively for the case of refreence implementation, and will let a student implement 'em as one desires?

lathspell: Is done!

lathspell:
At least the latest commit has no sneaked files according to `git show zebra-puzzle | diffstat`? Maybe my first push had them and I already deleted them.

lathspell:
I misinterpreted the diffstat output. Now they are back!

PR #400, opened 2019-12-20
Added new exercise 'yacht'

lathspell:
Another one


lathspell:
It's a good idea. I symlinked YachtCategory.kt.

PR #401, opened 2020-01-03
Added new exercise 'wordy'

lathspell:
Another one. Based on the Java example.


exercises/wordy/src/main/kotlin/Wordy.kt
```diff
@@ -0,0 +1,6 @@
+object Wordy {
+
+    fun solve(category: YachtCategory, vararg dices: Int): Int {
```
eparovyshnaya: should be something like 
```kotlin
fun asnwer (problem: String): Int
```

exercises/wordy/.meta/src/reference/kotlin/Wordy.kt
```diff
@@ -0,0 +1,43 @@
...
+                when (operator) {
+                    "plus" -> answerRecursive(rest, result + operand)
+                    "minus" -> answerRecursive(rest, result - operand)
+                    "multiplied by" -> answerRecursive(rest, result * operand)
+                    "divided by" -> answerRecursive(rest, result / operand)
+                    else -> throw IllegalArgumentException("Unknown operator '$operator'")
+                }
+            }
+            input.isEmpty() -> result
+            else -> throw IllegalArgumentException("Cannot parse '$input'?")
```
eparovyshnaya: Not a big deal, just wondering, why the exception message asks )

lathspell: Just a typo, it's now ending with an exclamation mark in both exceptions.

exercises/wordy/src/main/kotlin/Wordy.kt
lathspell: done

PR #403, opened 2020-01-05
Added new exercise 'resistor-color-duo'

lathspell:
Next one


exercises/resistor-color-duo/src/main/kotlin/ResistorColorDuo.kt
```diff
@@ -0,0 +1,6 @@
+object ResistorColorDuo {
+
+    fun colorCode(vararg input: String): Int {
```
eparovyshnaya: Let's discuss the signature.  
The problem description states that here we must implement a function that _decodes color code_ to a numeric _resistance value_. Such a query should most probably be named after the characteristic it queries, say 
```kotlin
fun resistance(vararg colors: String): Int
```
for domain-specific query, or
```kotlin
fun value(vararg colors: String): Int
```
for general-purpose one

What do you think?

exercises/resistor-color-duo/.meta/src/reference/kotlin/ResistorColorDuo.kt
```diff
@@ -0,0 +1,6 @@
+object ResistorColorDuo {
+
+    private val colors = listOf("black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white")
```
eparovyshnaya: Can we invent an **enum** here, please? )  It'd become a part of public api and that would follow the general habits of the track implementations.

lathspell: ok, can do :)
Should the tests use the constants from Colors enum or continue to call value() with strings?
The C track uses enum constants but defines them only in the example.h whereas the Java track uses string parameters and keeps Colors a `private enum` in the example solution.

eparovyshnaya: Here are lots of samples in our existing sourcebase, and you've recently implemented `Yacht` exercise with the same pattern (have a look at `YachtCategory` to recall):
 - define `enum class` in `src/main/kotlin` next to the stub implementation and thus this enum becomes a part of public api and can be used in signatures like 
```kotlin
 enum class Color ...

 fun value(vararg colors: Color): Int
```
 - place the same-named *link*-file next to reference implementaction 
 - use the enum both in ref implementation and test suite. I suppose, we must not insist on lower case, but should explore enum constants directly.

Does this sound reasonable?

lathspell: The challenge feels almost too easy now :)

Issue #360, opened 2019-09-23
Kotlin track - Forth issue


anwarsky:
1. No description provided
2. exercism download --exercise=forth --track=kotlin >> Error: failed to parse API error response: invalid character 'o' in literal null (expecting 'u')

blueglyph:
@anwarsky  That'd been reported in #302 already

Since the PR was abandoned, I found a work-around that could allow you to submit the solution, see the comment about 6 months ago, https://github.com/exercism/kotlin/issues/302#issuecomment-515762731

eparovyshnaya:
Closed by [PR 407](https://github.com/exercism/kotlin/pull/407)

Issue #302, opened 2019-05-27
Errors in: Forth


blueglyph:
The problem description is missing on the webpage (there is only the word 'Introduction').

When trying to download the problem, the following error occurs:

```
exercism download --exercise=forth --track=kotlin
Error: unable to parse API response - invalid character '<' looking for beginning of value
```

(I didn't have problems downloading the other tests)

OS: Win7 Pro x64

mikegehard:
Thanks for the report. 

I'm not sure why that even shows up on the site as it isn't in the config for the exercises:
https://github.com/exercism/kotlin/blob/master/config.json

nor is it in the exercises directory:
https://github.com/exercism/kotlin/tree/master/exercises

I've merged in some changes recently and see if those fix it. I'll continue to monitor.

blueglyph:
I've seen it in the branch kytrinyx/inline-insert, perhaps there was a good reason why it was removed.

Would you like me to have a look and test if that works? This seems to be quite an interesting test.

blueglyph:
I tried and solved it, it's indeed very interesting. I would probably add a couple of tests with corner cases (like a definition within a definition, empty definition ": ;" or strings that begin / end with spaces or are empty), but it's already very complete.

mikegehard:
Thanks for taking the time to figure this out. Is there a pull request that. I can take a look at?

blueglyph:
Maybe not, I added those few tests and they were all running fine on my solution, but I realize the reference version breaks with them, and I see no easy fix. It's an unnecessary addition anyway.

Taken as-is from the existing branch should be perfect.

Ric0chet-GitHub:
As of June 28th, this is still broken.  The situation described by blueglyph is still occurring.  It is not possible to start the "Forth" exercise.

> The problem description is missing on the webpage (there is only the word 'Introduction').
> 
> When trying to download the problem, the following error occurs:
> 
 ```
 exercism download --exercise=forth --track=kotlin
 Error: unable to parse API response - invalid character '<' looking for beginning of value
 ```

mikegehard:
Thanks for the report @Ric0chet-GitHub.

I haven't had any time to take a look due to other things going on. Not sure when I get to taking a look...sorry.

Ric0chet-GitHub:
Gotcha.  I realize there are only two Kotlin maintainers.  (It's frustrating as this is the only exercise I haven't completed in that track.)
  
[...]

PR #407, opened 2020-01-12
Added new exercise 'forth'

lathspell:
I was halfway through when I got the feeling of having done that before and noticed the open issues #302 and #360. Was forth accidently removed or did the original test really take to long? Anyway here is my version, in case you like it more than the original forth.
The test has inner classes because I played around with python/bin/generate_tests.py and a template modified for Kotlin.


exercises/forth/.meta/src/reference/kotlin/Forth.kt
```diff
@@ -0,0 +1,85 @@
...
+
+class Forth {
+
+    companion object {
+        private val NUMBER_PATTERN = Regex("""^-?\d+$""")
+    }
+
+    private var stack: Deque<Int> = ArrayDeque()
+    private var customOps: MutableMap<String, String> = mutableMapOf()
+    private var ops = mutableMapOf(
```
eparovyshnaya: Do you really need all these three fields declared as `var`s? There are no assignments for them, only referenced data structure mutations. They are effectively `val`s.

lathspell: Indeed, I changed them to vals.

PR #408, opened 2020-01-12
upgrade `list-ops` exercise and test suite to specification of 2.4.1version

eparovyshnaya:
 - ref implementation is fixed to satisfy new tests


uzilan:
Lgfm. Why were the hints removed though?

eparovyshnaya:
Restored, just in case it participates in the final exrcise presentation for a student.

Issue #291, opened 2019-03-11
Update simple-cipher to latest version of the cannonical data


mikegehard:
https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json

There was an initial attempt to update to 1.1.0 that I couldn't get to pass locally. If you want to start with that commit (https://github.com/exercism/kotlin/commit/d706824d7ef617aaa199bc53e026927e09d1cc42), revert it and then update to 2.0.0, that would be a good way to start.

eparovyshnaya:
Closed by #409

PR #409, opened 2020-01-13
#291 Update simple-cipher to latest version of the canonical data

eparovyshnaya:
upgrade `simple-cipher` test suite to specification of 2.0.0 version
 - test set is reworked according to the specification
 - ref implementation is fixed to satisfy the current requirements


uzilan:
LGFM

PR #416, opened 2020-01-26
Added new exercise 'scale-generator'

lathspell:
Another one. Didn't like the use of `var idx` but found no better solution.


exercises/scale-generator/src/test/kotlin/ScaleTest.kt
```diff
@@ -0,0 +1,109 @@
+import kotlin.test.assertEquals
+import kotlin.test.Test
+
+class ScaleTest {
+
+    // Test chromatic scales
+    @Test
+    fun `Chromatic scale with sharps`() {
```
eparovyshnaya: can we please start method name with lower case letter?

PR #417, opened 2020-02-20
binary-search-tree: add test case

glennj:
I see that some of the existing solution have not implemented the level-order algorithm correctly, but the mistakes are not exposed because the existing test cases are not deep enough. This change creates a tree 4 levels deep, and tests the asLevelOrderList() result.

The reference implementation correctly generates the level-order list of this bigger tree.


glennj:
I don't think so. The problem spec does not specify level-order traversal at all.

PR #419, opened 2020-04-10
kotlin/scale-generator: consider equivalent accidental notes to be equal. Fixes #418

### LINKED ISSUE #418: scale-generator: tests do not account for the fact that accidentals are equivalent
For example, in the problem description:

> Scales in Western music are based on the chromatic (12-note) scale. This scale can be expressed as the following group of pitches:
>
> A, A#, B, C, C#, D, D#, E, F, F#, G, G#
>
> A given sharp note (indicated by a #) can also be expressed as the flat of the note above it (indicated by a b) so the chromatic scale can also be written like this:
>
> A, Bb, B, C, Db, D, Eb, E, F, Gb, G, Ab

Logically, that should mean that we should accept either if we asked for the chromatic scale with tonic note "A". The same goes for any chromatic scale (or any series of intervals that includes accidentals).

However, the current iteration of the tests expects the results to be specifically the list specified in the problem-spec:  https://github.com/exercism/kotlin/blame/master/exercises/scale-generator/src/test/kotlin/ScaleTest.kt

Example:

```kotlin
val expected = listOf("C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B")
assertEquals(expected, Scale("C").chromatic())
```

Instead there should be a custom assertion that can check for equality of the accidentals:

```kotlin
import kotlin.test.asserter
...
assertScalesEqual(expected, Scale("C").chromatic())
...
fun assertScalesEqual(expected: List<String>, actual: List<String>) {
  // use kotlin.test.asserter to check for equality of the scales
  asserter.assertTrue(
    { messagePrefix("Scales should be equal") + "Expected <$expected>, actual <$actual>" },
    scalesAreEqual(expected, actual))
}

fun scalesAreEqual(expected: List<String>, actual: List<String>): Boolean {
    TODO("must allow that any accidental to be equivalent to the other version (eg. C# == Db)")
}
```

If I get the time, I can take a stab at implementing this.

jmrunkle:
Every sharp note has an equivalent flat note name (and vice versa). See #418 for more background.

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -2,6 +2,11 @@
 
 LATEST=https://github.com/exercism/configlet/releases/latest
 
+echo ">>> Fetching configlet..."
+echo
+echo "$ uname"
+uname
+
 OS=$(
 case $(uname) in
     (Darwin*)
@@ -24,6 +29,9 @@ case $OS in
         echo "tgz";;
 esac)
 
+echo "$ uname -m"
+uname -m
+
 ARCH=$(
 case $(uname -m) in
     (*64*)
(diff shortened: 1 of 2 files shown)


jmrunkle:
FWIW: I think that the failing CI test is a problem with the CI test itself and not the pull request.

PR #421, opened 2020-04-14
Add debug logging and fix fetch-configlet version to be case insensitive

### LINKED ISSUE #420: Continuous integration test bin/unit-test.sh is broken
I think this is also the case for exercism/java, but the CI test fails when it tries to download the configlet:

```
>>> Running configlet...
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
The command "$SCRIPT" exited with 2.
```

This is a problem with the `bin/unit-tests.sh` execution. More specifically, I think it is a problem in bin/fetch-configlet at [this line](https://github.com/exercism/kotlin/blob/master/bin/fetch-configlet#L49):

```
curl -s --location $URL | tar xz -C bin/;;
```

For whatever reason, the value returned does not apparently have a proper gzip format and it is breaking that integration test.

jmrunkle:
Fixes #420

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -2,6 +2,11 @@
 
 LATEST=https://github.com/exercism/configlet/releases/latest
 
+echo ">>> Fetching configlet..."
+echo
+echo "$ uname"
+uname
+
 OS=$(
 case $(uname) in
     (Darwin*)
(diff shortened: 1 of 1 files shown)


jmrunkle:
> Thanks for the fix!
> Just please, remove debug logging.

Actually, I want the debug logging to stick around for future issues. Also, this is exactly the file that we are using in exercism/java now (and I think we want them to match).

PR #463, opened 2021-04-20
CustomSet: Fix erroneous unit test expected value in CustomSetTest

jean-t86:
```
 @Test
  fun `union of non empty sets contains all unique elements`() {
      val set1 = CustomSet(1, 3)
      val set2 = CustomSet(2, 3)
      val expected = CustomSet(3, 2, 1)
      assertEquals(expected, set1 + set2)
  }
```

I think that this unit test is erroneous. The `expected` val should instead be `CustomSet(1, 2, 3)`

All the other unit tests keep the `CustomSet` sorted ascending except this one.

Without keeping the underlying data structure sorted in some way, the exercise cannot be done.

I will open a PR to fix this if that's alright.


jean-t86:
It was not a mistake in the test after having a closer look.

PR #469, opened 2021-05-09
Describe language features

dector:
Tell me if these are too generic - will try to bring other suggestions.

#468


ErikSchierboom (approved):
These key features are perfectly fine!

PR #470, opened 2021-05-09
Change linter command for CI

dector:
Seems like new configlet should be run as `configlet lint` but currently it's used as `configlet lint .`.

CI check will still fail but not because of Gradle. I'm checking it too (#471).


dector:
@ErikSchierboom not a big deal. I think we can remove linter from Travis CI later as we have it in GitHub Actions.

Issue #495, opened 2021-07-20
Update Kotlin to latest version


ErikSchierboom:
The exercises are using a somewhat outdated version of Kotlin: 1.3.60, whereas the latest version is 1.5.21.

Updating to the latest version will allow students to use all the latest and greatest features.

incendium:
This can be closed as #505 has resolved this issue.

SleeplessByte:
Thanks @incendium for the ping.

Issue #494, opened 2021-07-20
Update Gradle to latest version


ErikSchierboom:
When I open an exercise in IntelliJ, I get a warning message that some Gradle features are no longer supported. It would be nice to not have this warning by upgrading the Gradle version.

JonForest:
FWIW, we also get this in the command line when running the `./gradlew test` command
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.

incendium:
This can be closed as #505 has resolved this issue.

SleeplessByte:
Thanks @incendium for the ping.

PR #505, opened 2021-10-02
Update Kotlin and Gradle

incendium:
This PR is intended to tackle #494 and #495 as a first step towards helping reduce the build time issues wrt. exercism/kotlin-test-runner#30.

A summary of the changes:
* I restructured the build scripts a bit to both support the multi-module build used for testing all of the exercises via the reference implementations (the run-journey.sh script still remains broken, but running `./gradlew build` in `exercises` will work now).  All tests passed before I opened this PR.
* In order to support the weird project layout, I added a `settings.gradle.kts` file to each template to control the Kotlin version instead of directly defining it in the `build.gradle.kts` file.  This structure will play nicer with the Gradle KTS dialect in the multi-module project setup, and I have also manually verified that it works in standalone mode.  It's not quite ideal but it works.
* Gradle was updated to 7.3 and Kotlin was updated to 1.6.0.
* I added a (kind of hacky) script to fix the Git symlinks so that the exercise tests pass.  Some of the exercises use symlinks to the main source sets, and I could not get this to play well without manually fixing the symlinks.
* Ran some automated migrations to clean up warnings and avoid deprecated code slated for removal in future releases.
* Renamed test cases with `|` to use `-` instead as using `|` causes Windows systems to fail to compile the tests.

A future task is probably to update the JUnit tests to use v5 instead of v4, as I noticed some of the newer exercises are using JUnit 5.  This should be relatively easy to do with the vintage test runner for JUnit 5.


incendium:
Made the following changes:

* Gradle was updated to 7.3 and Kotlin was updated to 1.6.0.
* Ran some automated migrations to clean up warnings and avoid deprecated code slated for removal in future releases.
* Renamed test cases with | to use - instead as using | causes Windows systems to fail to compile the tests.

Thinking we're good to merge now.

SleeplessByte:
The `configlet` failure is unrelated to this PR:

```
[31m`/home/runner/work/kotlin/kotlin/concepts` contains a directory named `booleans`, which is not a `slug` in the concepts array. Please add the concept to that array:[0m
2021-11-20T22:36:31.8963106Z /home/runner/work/kotlin/kotlin/config.json
```

I am a bit wary that we don't have any CI that will determine if this change will break the track, but I think issues would be opened switfly if it does ;)

ErikSchierboom:
@incendium Great! Could you maybe also add a test case to the Kotlin test runner tests: https://github.com/exercism/kotlin-test-runner/tree/main/tests? That test case should use the new exercise structure (with the latest Gradle version).

incendium:
Sure, I will take a look at doing that when I get some free time.

PR #508, opened 2021-10-08
Ignoring all tests but the first one

uzilan:
For some reason, no skips were added to the tests even though this is not a concept exercise. I've seen several such exercises without skips and I think it will be better if they we're added, to allow TTD simulation (-ish).


uzilan:
If you think this is a valid change, please tag it with the hacktoberfest-accepted label. Thanks!

PR #514, opened 2021-11-27
Provides full testing

davidgaspardev:
Following the exercise description to provide a complete test.

### DIFF
--- a/exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
+++ b/exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
@@ -10,21 +10,18 @@ class TwoFerTest {
     }
 
     @Test
-    @Ignore
     fun aNameGiven() {
         assertEquals("One for Alice, one for me.", twofer("Alice"))
     }
 
     @Test
-    @Ignore
     fun anotherNameGiven() {
         assertEquals("One for Bob, one for me.", twofer("Bob"))
     }
 
     @Test
-    @Ignore
     fun emptyStringGiven() {
-        assertEquals("One for , one for me.", twofer(""))
+        assertEquals("One for you, one for me.", twofer(""))
     }
 
 }


exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
```diff
@@ -10,21 +10,18 @@ class TwoFerTest {
     }
 
     @Test
-    @Ignore
```
SleeplessByte, line 13: This should be ignored as in practice exericses, all tests are ignored except for the first.

exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
```diff
@@ -10,21 +10,18 @@ class TwoFerTest {
...
     fun aNameGiven() {
         assertEquals("One for Alice, one for me.", twofer("Alice"))
     }
 
     @Test
-    @Ignore
```
SleeplessByte, line 19: This should be ignored as in practice exericses, all tests are ignored except for the first.

exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
```diff
@@ -10,21 +10,18 @@ class TwoFerTest {
...
     fun anotherNameGiven() {
         assertEquals("One for Bob, one for me.", twofer("Bob"))
     }
 
     @Test
-    @Ignore
```
SleeplessByte, line 25: This should be ignored as in practice exericses, all tests are ignored except for the first.

exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
```diff
@@ -10,21 +10,18 @@ class TwoFerTest {
...
     fun emptyStringGiven() {
-        assertEquals("One for , one for me.", twofer(""))
+        assertEquals("One for you, one for me.", twofer(""))
```
SleeplessByte, line 24: Good catch, thanks!

SleeplessByte (changes requested):

kotp:
On 21/12/26 07:18PM, Derk-Jan Karrenbeld wrote:
> @SleeplessByte requested changes on this pull request.
> >      fun emptyStringGiven() {
> -        assertEquals("One for , one for me.", twofer(""))
> +        assertEquals("One for you, one for me.", twofer(""))
> 
> Good catch, thanks!

If a name is given it should be used.  If an empty string is given as
the name, then it is implied by "If not given a name, use 'you'" part of
the description (unless this has changed.)  If `nil` or `null` is given,
then that should be used, if `42` is given, then that should be used.

From: https://github.com/exercism/kotlin/blob/main/exercises/practice/two-fer/.docs/instructions.md

> Where "name" is the given name.
>
> However, if the name is missing, return the string:

The only time that '"you"' should be used, is when no name is given, at
least that is how I interpret this, and as far as I can see the existing
tests do not contradict this, though it also does not test either way,
which I find is a benefit for mentoring, rather than a weakness.

SleeplessByte:
Yeah, i don't mind keeping it for that reason. It's whatever you all want for mentoring 🙃😊.

If we were to follow Kotlin practices and make the argument optional, '' would not be the same as no string.

exercises/practice/two-fer/src/test/kotlin/TwoFerTest.kt
kotp: Let's keep this as it is, without the 'you' if truly we are giving an
empty string as a name.  (The giver of the name is the name expert,
after all, and we asked for it, after all.  Perhaps their parents named
them `""` or they legally changed their name to `""` (good luck with
that, but then Prince…))

SleeplessByte:
In that case, none of the changes are applicable.

kotp:
Thank you @davidgaspardev for bringing the proposed changes, as it helps to get some of the discussions out there as to why the code is the way it is.

PR #517, opened 2021-12-30
Added new exercise: Knapsack

### LINKED ISSUE #516: New exercise: Knapsack
I would like to implement the [knapsack](https://github.com/exercism/problem-specifications/tree/master/exercises/knapsack) exercise in this track.

lpizzinidev:
closes #516


SaschaMann:
@exercism/reviewers

exercises/practice/knapsack/src/main/kotlin/Knapsack.kt
```diff
@@ -0,0 +1,5 @@
+data class Item(val weight: Int, val value: Int)
+
+fun knapsack(maximumWeight: Int, items: List<Item>): Int {
+    TODO("Implement this function to complete the task")
+}
```
kotp, line 5: Missing EOL at what would be line 5.

kotp:
Other than that, it seems reasonable for someone that does not do Kotlin.  I do wonder if the jar file is something that is generated, and so should not be stored in the repository.

SleeplessByte (approved):

SleeplessByte:
@kotp the jar is indeed generated, but it's normal/common practice to commit it.

PR #540, opened 2022-07-10
Update HelloWorld.kt

Noam-Modai:


dector:
Hi, @Noam-Modai . The goal of this exercise is to return correct value from the function.

Seems like you didn't - congrats! 💪🏼 

There is no need to merge it back to the repository - feel free to play with code on you local machine.
You can also create your own repo for exercism solutions to keep track of them.

Good luck!

PR #543, opened 2022-08-23
Update HelloWorld.kt

Andres2310:


IndiGan (commented):
Hello, I think **you should close this pull request**, because the only thing you want to change is giving the solution directly.
But the goal is to find that we have to correct "Goodbye, Mars!" to "Hello, World!" instead of directly having "Hello, World!" and a 100% correct code.

SleeplessByte:
@IndiGan is correct. Please submit your solution to https://exercism.org via the web editor or the CLI (instructions in the hello world exercise on https://exercism.org)

PR #557, opened 2023-01-02
update gradle and junit

lsoares:


SleeplessByte:
Hi there @lsoares,

it seems that you're trying to upgrade the `forth` exercise from JUnit to the newer internal Kotlin test library, as well as upgrade `gradle`, if I am correct?

- Could you list some of the benefits?
- If approved and merged, will you apply the same to all the other exercises?
- Are there potential downsides to this change?
- Is the built-in Kotlin test suite now standard in "a lot" of Kotlin communities?

Normally we'd discuss changes like this first in issues (now: forum), which is why this PR was automatically closed.

lsoares:
Hey @SleeplessByte 

I don't mind discussing this in the forum. Should I?

To answer your questions:
In fact, I'm trying to update JUnit 4 to JUnit 5 - the way I'm doing this is through Kotlin but it's still an upgrade of JUnit.
JUnit 5 was released 6 years ago and JUnit 4 is being slowly deprecated (not sure when).
The benefits of JUnit 5 over JUnit 4 are a lot but for Exercism there are two that are worth it:
- Parameterized tests are way much easier and more readable ([more](https://www.baeldung.com/parameterized-tests-junit-5))
- Exception assertions are way better - no global rules or annotations - just use assertFails with a lambda (this one applies to this exercise and that's why I did it)

Regarding the upgrade: in 90% of the cases, it's just upgrading the imports. the remaining 10% is a bit more tricky (I'm referring to the advantages above as they were revamped features)
Yes, I can slowly upgrade the other exercises. However, there's a little "hack" that makes the test runner support both JUnit versions, so we don't need to rush (or even convert all if there are no clear/urgent benefits).

BTW, a question: why is the runner not available from the website?

JUnit 5 is the standard now. Nobody uses 4 in a new project (recall that Kotlin is just wrapping it so that's a detail - we could have imported JUnit 5 directly).
I don't see any downside because the upgrade can be done optionally and per exercise.

SleeplessByte:
> BTW, a question: why is the runner not available from the website?

The runner would always timeout because gradle startup and build is too slow. The entire execution must be under ten seconds and we were not yet able to get it that low.

SleeplessByte:
Can you explain how this change makes it easier to run the tests from the IDE? Documenting that here, in this PR, can aid with future discussions about the topic :)

lsoares:
with JUnit 5 nested abilities:
image

image

with JUnit 4 there's no arrow to run the test and it's not visible in the menu as well.
image


PS. I doubt we can run Gradle under 10 secs unless we keep Gradle cache.

PR #566, opened 2023-03-09
flatten-array: test for heterogeneous lists

gypsydave5:
Hi 👋 

This adds a test to the `flatten-array` problem that forces solutions to work on `Collection<Any?>`. At present they can get away with just `Collection<Int?>`.

Nothing too controvertial (I hope).

### DIFF
--- a/exercises/practice/flatten-array/src/test/kotlin/FlattenerTest.kt
+++ b/exercises/practice/flatten-array/src/test/kotlin/FlattenerTest.kt
@@ -47,7 +47,7 @@ class FlattenerTest {
 
     @Ignore
     @Test
-    fun flattensHetrogenousList() {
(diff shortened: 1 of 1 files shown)


exercises/practice/flatten-array/src/test/kotlin/FlattenerTest.kt
```diff
@@ -45,4 +45,10 @@ class FlattenerTest {
         assertEquals(emptyList<Any>(), Flattener.flatten(nestedList))
     }
 
+    @Ignore
+    @Test
+    fun flattensHetrogenousList() {
```
ErikSchierboom, line 50:
```suggestion
    fun flattensHeterogenousList() {
```

PR #577, opened 2023-03-27
Raindrops approaches

bobahop:
Approaches for Raindrops

### DIFF
--- a/exercises/practice/raindrops/.approaches/buildstring/content.md
+++ b/exercises/practice/raindrops/.approaches/buildstring/content.md
@@ -1,0 +1,41 @@
+# `buildString`
+
+```kotlin
+object Raindrops {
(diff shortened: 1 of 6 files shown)


exercises/practice/raindrops/.approaches/buildstring/content.md
```diff
@@ -0,0 +1,41 @@
+# `buildString`
+
+```kotlin
+object Raindrops {
+    fun convert(num: Int) = buildString {
+        if (num % 3 == 0) append("Pling")
+        if (num % 5 == 0) append("Plang")
+        if (num % 7 == 0) append("Plong")
+        if (isEmpty()) append(num)
+    }
```
ErikSchierboom, line 10: So nice!

PR #592, opened 2023-05-19
Update Beer-Song empty solution to have return type

aljidy:
I believe this may lead to less confusion and makes a bit more sense as the Unit tests expect a string, so leaving it out doesn't exactly lend itself to be more challenging, just unclear.


SleeplessByte (approved):
Sure!

PR #593, opened 2023-05-19
Update readme - gradle test instructions

aljidy:
Updated the readme instructions for how to run tests against an exercise like `bob` as it wasn't clear that it was in the practice folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ To run the tests for a specific exercise, run the `test` Gradle task from the ex
 directory. For example:
 
 ```bash
-cd exercises
+cd exercises/practice
 https://github.com/exercism/v3/blob/main/gradlew bob:test
 
 ```


README.md
```diff
@@ -91,7 +91,7 @@ To run the tests for a specific exercise, run the `test` Gradle task from the ex
 directory. For example:
```
SleeplessByte, line 91: It should have picked up the exercise from the `exercises` directory.

aljidy: I might have missed something but the exercises (like `bob`) aren't in the root but in the `practice` subfolder, please see the below that represents the current structure

```
.
├── ...
├── README.md
├── exercises
│  
│   ├── concept
│   │   ├── annalyns-infiltration
│   │   └── lucians-luscious-lasagna
│   ├── practice
│   │   ├── accumulate
│   │   ├── beer-song
│   │   ├── bob
                            ...
│   │   └── zebra-puzzle
...
```

SleeplessByte: Yes, that's correct, but the `gradlew` executable generally searches multiple levels deep.

`bob:test` isn't "folder bob, run test", it's find the `bob` project and run the `test` task. `bob` _should_ be a [subproject](https://github.com/exercism/kotlin/blob/main/exercises/build.gradle.kts#L7) of the `practices` subproject.

Note that you may be completely right that the command needs to be `exercises/practices`. I am just commenting on the current status quo how the instructions are written. If that doesn't work as advertised, we MUST update the code that is supposed to run the tests or update the documentation as you suggested.

I would recommend opening a forum post so it can be discussed in public and then we come back here and apply the necessary changes, if any.

aljidy: Thank you so much for clarifying, I'll give it another try and if it's not working as expected, I'll look through the Gradle tasks/files!

PR #604, opened 2023-09-28
configlet fmt for approach configs

BNAndras:


BNAndras:
@exercism/kotlin, no additional configs need to be formatted.

PR #617, opened 2024-02-01
Add exercise: protein-translation 🧬 

AnneKlapwijk:
## Link to forum posts

- https://forum.exercism.org/t/kotlin-protein-translation-exercise-48in24/9605
- https://forum.exercism.org/t/48in24-exercise-02-13-protein-translation/9556

## What has been done
- Use `configlet` to generate common files
- Update some config files
- Implement test cases (got the idea for a parameterized test from the JS track: https://github.com/exercism/javascript/blob/main/exercises/practice/protein-translation/protein-translation.spec.js#L22)
- Write stub implementation: provide a starting point for students
- Write an example solution that passes all the tests
  - `./gradlew practice:protein-translation:test`

## Notes
First PR here. Let me know if I missed anything 😊


AnneKlapwijk:
Ping @exercism-kotlin 🙂 🛎️

exercises/settings.gradle.kts
```diff
@@ -53,6 +53,7 @@ include(
     "practice:pangram",
     "practice:pascals-triangle",
     "practice:perfect-numbers",
+    "practice:protein-translation",
```
SleeplessByte, line 56: This list looks lexicographical, so could you move this below `"practice:prime-factors"`, so the order is preserved?

SleeplessByte (changes requested):
Maar 1 kleine opmerking

exercises/settings.gradle.kts
AnneKlapwijk: Well spotted! Thanks for reviewing! 😊

SleeplessByte (approved):

PR #629, opened 2024-04-01
Kotlin / Bob / DigDeeper: Add a pattern matching approach using when-expression

micha-b:
Hey there,

as proposed in the forum [here](https://forum.exercism.org/t/kotlin-bob-digdeeper-maybe-add-a-pattern-matching-approach/10223) I added a new approach to the Bob exercise on the Kotlin track using a `when`-expression.

I hope I provided all the files and changes needed for a correctly working version.

I am not quite sure if the entry in the config.json file needs to be done manually (as I did) or if this entry is somehow generated automatically. (I used a randomly generated uuid for the entry.)

Please have a look at the proposed changes and give feedback on the writing and configuration.

Have a nice day! 👋🏼

### DIFF
--- a/exercises/practice/bob/.approaches/when-expression/content.md
+++ b/exercises/practice/bob/.approaches/when-expression/content.md
@@ -1,0 +1,60 @@
+# `when` expressions
+
+```kotlin
+object Bob {
+    fun hey(statement: String): String =
+        when {
+            statement.isQuestion() && statement.isYelling() -> "Calm down, I know what I'm doing!"
+            statement.isQuestion() -> "Sure."
+            statement.isYelling() -> "Whoa, chill out!"
+            statement.isSilence() -> "Fine. Be that way!"
+            else -> "Whatever."
+        }
+
+    private fun String.isSilence(): Boolean = this.isBlank()
+    private fun String.isQuestion(): Boolean = this.trim().endsWith('?')
+    private fun String.isYelling(): Boolean {
+        val letters = this.filter { it.isLetter() }
+        return if (letters.isEmpty())
+            false
+        else
+            letters.all { it.isUpperCase() }
+    }
+}
+```
+
+In this approach you have a `when` expression containing different so-called branches that can be matched using the corresponding patterns. 
+As soon as one of these patterns on the left side of the arrow (`->`) is matched ...
+
+1. the value on the right side
+2. the block on the right side is executed and the value retuned from the block 
+
+... is returned from the `when` expression.
+
+Only one branch is matched in one execution of the `when` expression. The branches are matched from top to bottom and the first branch in which the condition evaluates to `true` is selected. 
+If none of the given conditions matches the `else` branch is selected and returned.
+
+~~~~exercism/caution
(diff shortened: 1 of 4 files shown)


exercises/practice/bob/.approaches/when-expression/snippet.txt
ErikSchierboom, line 1: The snippet can be at most 8 lines. You should try to pick the lines that best represent the approach (so that would be the `when` expression, with the function definition I think).

exercises/practice/bob/.approaches/when-expression/content.md
```diff
@@ -0,0 +1,60 @@
...
+    private fun String.isSilence(): Boolean = this.isBlank()
+    private fun String.isQuestion(): Boolean = this.trim().endsWith('?')
+    private fun String.isYelling(): Boolean {
+        val letters = this.filter { it.isLetter() }
+        return if (letters.isEmpty())
+            false
+        else
+            letters.all { it.isUpperCase() }
+    }
+}
```
ErikSchierboom, line 23: What about?

```suggestion
    private fun String.isYelling() = any(Char::isLetter) && toUpperCase() == this
```

exercises/practice/bob/.approaches/when-expression/content.md
```diff
@@ -0,0 +1,60 @@
...
+        when {
+            statement.isQuestion() && statement.isYelling() -> "Calm down, I know what I'm doing!"
+            statement.isQuestion() -> "Sure."
+            statement.isYelling() -> "Whoa, chill out!"
+            statement.isSilence() -> "Fine. Be that way!"
+            else -> "Whatever."
+        }
+
+    private fun String.isSilence(): Boolean = this.isBlank()
+    private fun String.isQuestion(): Boolean = this.trim().endsWith('?')
```
ErikSchierboom, line 15: The return types could be omitted (feel free to ignore this suggestion)

```suggestion
    private fun String.isSilence() = this.isBlank()
    private fun String.isQuestion() = this.trim().endsWith('?')
```

ErikSchierboom (commented):
> I am not quite sure if the entry in the config.json file needs to be done manually (as I did) or if this entry is somehow generated automatically. (I used a randomly generated uuid for the entry.)

Both are fine. You could generate it via:

```
bin/fetch-configlet
bin/configlet create --approach when-expression --exercise bob
```

exercises/practice/bob/.approaches/when-expression/content.md
micha-b: Hey, 
looks good. Tried a solution using higher order functions but did not find an approach I was happy with.

I will update the solution using this one-liner, thanks.

exercises/practice/bob/.approaches/when-expression/content.md
micha-b: I usually really like explicit return types because I think it makes the intent and outcome of a function easier to read.

But I know that you are able to omit them in Kotlin if you want to. 
But I usually don't omit them.

ErikSchierboom: Okay, that's perfectly fine :)

ErikSchierboom (approved):
Thanks! Looks good.

PR #703, opened 2025-06-30
Add `flower-field`

BNAndras:
https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/17967


SleeplessByte (approved):
Awesome!

PR #710, opened 2025-08-29
Add Conditionals concept

colinleach:
The first new concept for several years is short and boring. That seems appropriate to the topic, though it is still an important early concept.

The `introduction.md` is provisional. I always assume these will be refined when adding the Concept Exercise, and that version will then be copied back to the Concept. (Bethany taught me that, and it worked well on the recent Julia syllabus)

### DIFF
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -1,0 +1,69 @@
+# About conditionals
+
+## Comparison operators
+
+[Comparison operators][operators] are similar to many other languages, with a few extensions.
+
+For equality, the operators are `==` (equal) and `!=` (not equal).
+
+```Kotlin
+val txt = "abc"
+txt == "abc" // => true
+txt != "abc" // => false
+```
+
+Additionally, the `===` and `!==` operators test for ["referential equality"][referential-equality]:
+`a === b` if and only if `a` and `b` point to the same object.
+This should make more sense later in the syllabus.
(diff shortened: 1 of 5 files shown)


colinleach:
@SleeplessByte (no rush, have a good weekend)

concepts/conditionals/about.md
```diff
@@ -0,0 +1,69 @@
...
+- Both `else if` and `else` are optional, and there can be multiple `else if` blocks.
+
+
+## Alternatives?
+
+By deliberate choice, Kotlin does _not_ have the ternary operator `? :` found in Java.
+A concise form of `if ... else` is preferred:
+
+```Kotlin
+return if (isOK) goodValue else badValue
```
SleeplessByte, line 55: Maybe it makes sense to have two examples:

```kotlin
val result = if (isOk) goodValue else badValue
```

(and the one you have).

And maybe:

````
Unlike Ruby, the concise `if ... else` form *always* needs an `else` and thus cannot be used as a guard statement:

```kotlin
return 42 if (isOk)
// Syntax error: Unexpected tokens (use ';' to separate expressions on the same line).
// Syntax error: Expecting an expression.

return if (isOk) true
// 'if' must have both main and 'else' branches when used as an expression.
```

What do you think?

SleeplessByte (approved):

PR #711, opened 2025-08-30
Add Numbers concept

colinleach:
The `introduction.md` is too long, but we need to pick a concept exercise before deciding what is needed and what can be deleted.

### DIFF
--- a/concepts/numbers/about.md
+++ b/concepts/numbers/about.md
@@ -1,0 +1,148 @@
+# About Numbers
+
+## Numerical types
+
+[Numbers][numbers] can be integer, unsigned integer, or floating point types.
+Each comes in various "sizes", meaning how many bits it needs in memory.
+
+- Integers can be `Byte`, `Short`, `Int` or `Long`, respectively 8, 16, 32 and 64 bits.
+- Unsigned integers have a `U` prefix: `UByte`, `UShort`, `UInt` or `ULong`.
+- Floating point types are `Float` (32-bit) or `Double` (64-bit).
+
+Integer variables relying on type inference default to `Int`, even on 64-bit machines, but floating point variables default to `Double`.
+
+Other types can, of course, be specified, but there are a few syntactic shortcuts, and big integer literals become `Long` if they would overflow `Int`.
+
+```Kotlin
+val one = 1 // defaults to Int
+val threeBillion = 3_000_000_000 // Long, with optional underscores for clarity
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+val oneDouble = 1.0 // defaults to Double
+val oneFloat = 1.0f //Float
+val lightSpeed = 3.0e8 // scientific notation (units of m/s)
+```
+
+Hexadecimal and binary literals are conventional: `0x7F` and `0b100101` respectively.
+Octal literals are not supported in Kotlin.
+
+## Arithmetic
+
+The basic arithmetic operators are the same as in many languages:
+
+```Kotlin
+4 + 3 // => 7
+4 - 3 // => 1
+4 * 3 // => 12
+4 / 3 // => 1  Int / Int always gives an Int
+-8 / 3 // => -2  Truncated towards zero
+-8.0 / 3 // => -2.6666666666666665
+```
+
+To get a floating point result from division, at least one of the numerator / denominator must be floating point.
+
+Division by zero is more interesting.
+
+```Kotlin
+1 / 0 // => java.lang.ArithmeticException: / by zero
+3.0 / 0.0 // => Infinity
+0.0 / 0.0 // => NaN (Not a Number)
+```
+
+Integer division by zero is an error, but [IEEE floating point standards][IEEE] can apply in other cases.
+
+The modulo operator `%` gives the remainder from integer division:
+
+```Kotlin
+8 % 3 // => 2
+```
+
+Kotlin, like other JVM languages, has no exponentiation operator (_this annoys scientists and engineers_).
+We need to use the [`pow()`][pow] function from the [`math`][math] library, and the number being raised to some power must be `Float` or `Double`.
+
+```kotlin
+2.0.pow(3)  // => 8.0
+2.pow(3)  // Unresolved reference (2 is Int, so not allowed)
+```
+
+## Rounding
+
+The [`math`][math] library contains several functions to round floating point numbers to a nearby integer.
+
+_Did the last line sound slightly odd?_
+When we say "a nearby integer", there are two questions:
+
+- Is the return value still Float/Double, or is it converted to Int/Long?
+- How are ties rounded? Does `4.5` round to `4` or `5`?
+
+It is not Kotlin's fault if this seems complicated.
+Mathematicians have been arguing about this for centuries.
+
+### `round()`, `floor()`, `ceil()`, `truncate()`
+
+These four functions all return the same floating-point type as the input, but differ in how they round.
+
+- [`round()`][round] gives the _nearest_ integer if this is unambiguous, or the _nearest even_ integer when tied.
+- [`floor()`][floor] rounds towards negative infinity.
+- [`ceil()`][ceil] rounds towards positive infinity
+- [`truncate()`][truncate] rounds towards zero
+
+```kotlin
+round(4.7)  // => 5.0 (nearest integer)
+round(4.5)  // => 4.0 (nearest even integer)
+
+floor(4.7)  // => 4.0  (towards -Inf)
+floor(-4.7)  // => 5.0
+
+ceil(4.7)   // => 5.0  (towards +Inf)
+ceil(-4.7)  // => -4.0
+
+truncate(4.7)  // => 4.0  (towards zero)
+truncate(-4.7) // => -4.0
+```
+
+### `roundToInt()`, `roundToLong()`
+
+These two functions do a type conversion after rounding, so the return type is `Int` or `Long` respectively.
+
+Ties are always rounded towards positive infinity.
+
+```kotlin
+4.3.roundToInt()  // => 4 (nearest integer)
+4.5.roundToInt()  // => 5  (nearest integer towards +Inf)
+(-4.5).roundToInt()  // => -4
+```
+
+Note also the different syntax: `x.roundToInt()` versus `round(x)`.
+
+## Type conversions
+
+Kotlin will quietly do implicit conversions in a few cases, for example `Int` to `Double` in a mixed arithmetic expression:
+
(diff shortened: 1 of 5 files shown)


concepts/numbers/about.md
```diff
@@ -0,0 +1,148 @@
...
+[Numbers][numbers] can be integer, unsigned integer, or floating point types.
+Each comes in various "sizes", meaning how many bits it needs in memory.
+
+- Integers can be `Byte`, `Short`, `Int` or `Long`, respectively 8, 16, 32 and 64 bits.
+- Unsigned integers have a `U` prefix: `UByte`, `UShort`, `UInt` or `ULong`.
+- Floating point types are `Float` (32-bit) or `Double` (64-bit).
+
+Integer variables relying on type inference default to `Int`, even on 64-bit machines, but floating point variables default to `Double`.
+
+Other types can, of course, be specified, but there are a few syntactic shortcuts, and big integer literals become `Long` if they would overflow `Int`.
```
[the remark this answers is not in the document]
colinleach: I've added a paragraph - is this what you wanted?

concepts/numbers/introduction.md
```diff
@@ -0,0 +1,140 @@
...
+Kotlin, like other JVM languages, has no exponentiation operator.
+We need to use the `pow()` function from the [`math`][math] library, and the number being raised to some power must be `Float` or `Double`.
+
+```kotlin
+2.0.pow(3)  // => 8.0
+2.pow(3)  // Unresolved reference (2 is Int, so not allowed)
+```
+
+## Rounding
+
```
kahgoh, line 69: @colinleach I wasn't quite sure if you meant to remove bits from this introduction as well or just from the exercise's introduction in the [comment](https://github.com/exercism/kotlin/pull/713#issuecomment-3240268568) (ignore this if you meant just from the exercise introduction), but if you're looking for something to cut or separate out, the Cars Assemble exercise dosen't cover rounding and the use of the math library.

concepts/numbers/introduction.md
```diff
@@ -0,0 +1,140 @@
...
+```Kotlin
+4 + 3 // => 7
+4 - 3 // => 1
+4 * 3 // => 12
+4 / 3 // => 1  Int / Int always gives an Int
+-8 / 3 // => -2  Truncated towards zero
+-8.0 / 3 // => -2.6666666666666665
+```
+
+To get a floating point result from division, at least one of the numerator / denominator must be floating point.
```
kahgoh, line 42: Isn't this true for any arithmetic operator in Kotlin? For example `3 + 2.0` will give `5.0` and `3 * 2.0` gives `6.0` (both results are doubles).

concepts/numbers/introduction.md
```diff
@@ -0,0 +1,140 @@
...
+val one = 1 // defaults to Int
+val threeBillion = 3_000_000_000 // Long, with optional underscores for clarity
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+val oneDouble = 1.0 // defaults to Double
+val oneFloat = 1.0f //Float
+val lightSpeed = 3.0e8 // scientific notation (units of m/s)
+```
+
+Hexadecimal and binary literals are conventional: `0x7F` and `0b100101` respectively.
```
kahgoh, line 26: Similar to another comment, hexadecimal, binary and octal aren't currently covered by Cars Assemble either.

concepts/numbers/introduction.md
colinleach: If you make a cut-down version in the exercise, I'll wait till that is merged then copy it back to the concept. Probably easier than trying to keep them in sync as yours goes through review and editing.

concepts/numbers/introduction.md
colinleach: No problem, just cut it out of your introduction along we everything else you don't need. Deleting is quicker than creating!

I'll keep it in the About, which aims to be a reference document with more complete information.

concepts/numbers/about.md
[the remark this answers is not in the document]
SleeplessByte: It's better, but assume someone doesn't know what "number can be x bytes" means. I am a strong opponent to "just" explain this, especially in `about.md`, but also in `introduction.md` if it's relevant. We could add above line 16 something like:

```
In Kotlin, like many strongly types languages, numbers take a fixed amount of space in memory, which also means for each numeric type there is a smallest and largest value that fits. This is generally expressed in bits or bytes.

For numeric types that can go below and above zero, one bit is reserved to indicate the *sign*. This effectively cuts the amount of integers that fits in the memory in half, and these are called signed integers. In contrast, there are also unsigned integers which do not reserve a bit for the sign, doubling the maximum value the memory for that number can hold.
```

You may want to reword in your own words.

Sidenote: if you feel my or `kahgoh`'s contributions are significant, feel free to add us to the contributors array when you see fit.

concepts/numbers/about.md
```diff
@@ -0,0 +1,160 @@
...
+val one = 1 // defaults to Int
+val threeBillion = 3_000_000_000 // Long, with optional underscores for clarity
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+val oneDouble = 1.0 // defaults to Double
+val oneFloat = 1.0f // Float
+val lightSpeed = 3.0e8 // scientific notation (units of m/s)
+```
+
+Hexadecimal and binary literals are the same as in many languages: `0x7F` and `0b100101` respectively.
```
SleeplessByte, line 34:
```suggestion
Hexadecimal literals start with `0x`, e.g. `0x7F`, and binary literals start with `0b`, e.g. `0b100101`.
```

concepts/numbers/about.md
```diff
@@ -0,0 +1,160 @@
...
+
+The basic arithmetic operators are the same as in many languages:
+
+```Kotlin
+4 + 3 // => 7
+4 - 3 // => 1
+4 * 3 // => 12
+4 / 3 // => 1  Int / Int always gives an Int
+-8 / 3 // => -2  Truncated towards zero
+-8.0 / 3 // => -2.6666666666666665
```
SleeplessByte, line 47:
````suggestion
12 / 4 => 3
```

When dividing an integer (for example an Int) by another integer (for example an Int), the result will also be an integer (in this case an Int):

```kotlin
4 / 3 // => 1  Int / Int always gives an Int
-8 / 3 // => -2  Truncated towards zero
```

concepts/numbers/about.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+To get a floating point result, at least one of the numbers in the calculation must be floating point.
+This is particularly significant for division, as integer division often leads to truncation.
+
```
SleeplessByte, line 52:
````suggestion
To get a floating point result, at least one of the numbers in the calculation must be floating point (ie. a Float or a Double).
This is particularly significant for division, as integer division often leads to truncation.

```kotlin
-8.0 / 3   // => -2.6666666666666665
-8 / 3.0   // => -2.6666666666666665
-8.0 / 3.0 // => -2.6666666666666665
```
````

File: concepts/numbers/about.md [hunk 8]
```diff
@@ -0,0 +1,160 @@
...
+Division by zero is more interesting.
+
+```Kotlin
+1 / 0 // => java.lang.ArithmeticException: / by zero
+3.0 / 0.0 // => Infinity
+0.0 / 0.0 // => NaN (Not a Number)
+```
```
SleeplessByte, line 59:
````suggestion
```Kotlin
1 / 0     // => java.lang.ArithmeticException: / by zero
3.0 / 0.0 // => Infinity
0.0 / 0.0 // => NaN (Not a Number)
```
````

This may be easier to read.

You can also add the reason for this:

````suggestion
```Kotlin
1 / 0     // => java.lang.ArithmeticException: / by zero
3.0 / 0.0 // => Infinity
0.0 / 0.0 // => NaN (Not a Number)
```

That's because floating point operations follow the [IEEE-standard for Floating Point math](https://en.wikipedia.org/wiki/IEEE_754).
````

File: concepts/numbers/about.md [hunk 9]
```diff
@@ -0,0 +1,160 @@
...
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+val oneDouble = 1.0 // defaults to Double
+val oneFloat = 1.0f // Float
+val lightSpeed = 3.0e8 // scientific notation (units of m/s)
+```
+
+Hexadecimal and binary literals are the same as in many languages: `0x7F` and `0b100101` respectively.
+Octal literals are not supported in Kotlin.
+
```
SleeplessByte, line 36: Maybe also add the Kotlin does not support arbitrary precision arithmetic in its standard library, so a userland solution is necessary for `BigInteger` and `BigDecimal` support?

SleeplessByte (approved):

concepts/numbers/about.md
```diff
@@ -0,0 +1,160 @@
...
+4 + 3 // => 7
+4 - 3 // => 1
+4 * 3 // => 12
+4 / 3 // => 1  Int / Int always gives an Int
+-8 / 3 // => -2  Truncated towards zero
+-8.0 / 3 // => -2.6666666666666665
+```
+
+To get a floating point result, at least one of the numbers in the calculation must be floating point.
+This is particularly significant for division, as integer division often leads to truncation.
```
kahgoh, line 51: Suggestion for explaining Kotlin converts types in calculation. 

````suggestion
The basic arithmetic operators are the same as in many languages.
When the numbers in the calculation are the same type, the result will also be the same type:

```kotlin
4 + 3       // => 7     Int + Int = Int
4 - 3       // => 1     Int - Int = Int
4L * 3L     // => 12    Long * Long = Long
6.0 / 2.0   // => 3.0   Double / Double = Double
```

If the numbers are different types, Kotlin will infer the result's type to mitigate data loss:

```kotlin
4 + 3.0         // => 7.0                   Int + Double = Double
10.5 - 3.125f   // => 7.375                 Double - Float = Double
7L * 1.125      // => 7.875                 Long * Double = Double
10L / 3.0       // => 3.3333333333333335    Long / Double = Double
```

This is particularly significant for division, as dividing an integer (for example an Int) by another integer (for example an Int) will give an integer (in this case an Int):

```kotlin
4 / 3   // => 1     Int / Int always gives an Int
-8 / 3  // => -2    Truncated towards zero
```

To keep the decimal part, convert one of the numbers to a Float or Double:

```kotlin
-8.0 / 3    // => -2.6666666666666665
-8 / 3.0    // => -2.6666666666666665
-8.0 / 3.0  // => -2.6666666666666665
```

````

PR #712, opened 2025-08-30
Add Strings concept

colinleach:
Again, there is too much detail in `introduction.md`, so it will need pruning later to fit the exercise.

### DIFF
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -70,7 +70,8 @@ Mostly, these are [`extensions functions`][ref-extensions] rather than members o
 
 ~~~~exercism/note
 Kotlin's rather complex [documentation][ref-string-functions] pages hide extension functions in the default view.
-Be sure to click `Members and Extensions` to expand this section.
+At moment of writing this, the most valuable content is hidden in a tab named `Members and Extensions`.
+Click it to expand this section and see all the members and extensions available on the `String` class.
 
 [ref-string-functions]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/
 ~~~~
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -68,8 +68,12 @@ Kotlin provides _many_ [`functions`][ref-string-functions] to manipulate strings
 
 Mostly, these are [`extensions functions`][ref-extensions] rather than members of the `String` class, though this has little effect on how we use them.
 
-***Note:*** _Kotlin's rather complex [documentation][ref-string-functions] pages hide extension functions in the default view.
-Be sure to click `Members and Extensions` to expand this section._
+~~~~exercism/note
+Kotlin's rather complex [documentation][ref-string-functions] pages hide extension functions in the default view.
+Be sure to click `Members and Extensions` to expand this section.
+
+[ref-string-functions]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/
+~~~~
 
 The following example shows just a small selection of what is available:
 
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -1,0 +1,148 @@
+# About Strings
+
+A [`string`][string] in Kotlin is an immutable sequence of Unicode characters.
+
+[`Immutable`][immutable] means that any operation on a string must return a new string: the original string can never change.
+
+[`Unicode`][unicode] means that most of the world's writing systems can be represented, but (in contrast to older languages such as C) there is no 1:1 mapping between characters and bytes.
+This will be discussed further in the [`Chars`][chars] Concept.
+
+A string is surrounded by double-quotes `" "`.
+
+Some characters need escaping: `\'`, `\\`, plus the usual non-printing characters such as `\t` (tab) and `\n` (newline).
+
+```kotlin
+val s = "Escape apostrophe \' and backslash \\."
+// Escape apostrophe ' and backslash \.
+```
+
+Raw strings use 3 double-quotes, and can contain arbitrary text (no need for escaping).
+Multiline strings are also supported, including flexible handling of indents.
+
+```kotlin
+val multi = """I'm a
+    |multi-line
+    |string with special characters \ \t """
+
+multi.trimMargin()  // delimiter defaults to | but can be specified
+//I'm a
+//multi-line
+//string with special characters  \ \t 
+```
+
+Strings can be concatenated with `+`, but this is best limited to short and simple cases.
+There are other and often better options.
+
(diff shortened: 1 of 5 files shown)


concepts/strings/about.md
```diff
@@ -0,0 +1,148 @@
...
+
+[`Unicode`][unicode] means that most of the world's writing systems can be represented, but (in contrast to older languages such as C) there is no 1:1 mapping between characters and bytes.
+This will be discussed further in the [`Chars`][chars] Concept.
+
+A string is surrounded by double-quotes `" "`.
+
+Some characters need escaping: `\'`, `\\`, plus the usual non-printing characters such as `\t` (tab) and `\n` (newline).
+
+```kotlin
+val s = "Escape apostrophe \' and backslash \\."
```
SleeplessByte, line 15: `'` doesn't need to be escaped!

```kotlin
println("Escape apostrophe ' and backslash \\.")
// => Escape apostrophe ' and backslash \.
```

concepts/strings/about.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+Strings can be concatenated with `+`, but this is best limited to short and simple cases.
+There are other and often better options.
+
+## String templates
+
+[`Templates`][templates] refers to what some other languages call "interpolation".
+
+if a dollar sign `$` is followed by an identifier or expression within a string, the _value_ is substituted.
```
SleeplessByte, line 40:
```suggestion
If a string contains a dollar sign `$`, followed by an identifier, or contains braces (`{expression}`) surrounding an expression, those are substituted by respectively the value or the result of the expression.
```

concepts/strings/about.md
```diff
@@ -0,0 +1,148 @@
...
+The braces `{ }` are needed around expressions when parsing would otherwise be ambiguous.
+
+## String formatting
+
+On the JVM platform (only), `String.format()` allows more precise formatting than string templates, with [syntax][formats] similar to the (_very old!_) [`printf`][printf] functions.
+
+```kotlin
+String.format("%s %.3f", "π ≈", 3.14159)
+//π ≈ 3.142
+```
```
SleeplessByte, line 57: Add 

````markdown
~~~exercism/advanced
// ...
~~~
```` 

Explaining in a few lines that Kotlin has different compile platforms with different capabilities.

File: concepts/strings/about.md [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+
+## String functions
+
+Kotlin provides _many_ [`functions`][string-functions] to manipulate strings.
+
+Mostly, these are [`extensions functions`][extensions] rather than members of the `String` class, though this has little effect on how we use them.
+
+***Note:*** _Kotlin's rather complex [documentation][string-functions] pages hide extension functions in the default view.
+Be sure to click `Members and Extensions` to expand this section._
```
SleeplessByte, line 66: Use an `exercism/...` admonition for increased differentiating from the normal content.

`Members and Extension` is specific to certain IDEs, list them!

concepts/strings/about.md
```diff
@@ -0,0 +1,148 @@
...
+[string-functions]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/
+[chars]: https://exercism.org/tracks/kotlin/concepts/chars
+[lists]: https://exercism.org/tracks/kotlin/concepts/lists
+[templates]: https://kotlinlang.org/docs/strings.html#string-templates
+[sb-append]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-string-builder/#9100522%2FFunctions%2F-705004581
+[sb-insert]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-string-builder/#-132863384%2FFunctions%2F-705004581
+[sb-deleteat]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-string-builder/#-386007892%2FFunctions%2F-956074838
+[sb-deleterange]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-string-builder/#-1622040372%2FFunctions%2F-956074838
+[buildstring]: https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#build-a-string
+[jointostring]: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/join-to-string.html
```
SleeplessByte, line 148: For ease of mainting, I prefer:

- reference-string or ref-string instead of string
- wiki-unicode instead of unicode
- web-formats instead of formats
- concept-chars instead of chars

It makes it easy to see the "type" of resource without finding the reference here.

concepts/strings/about.md
colinleach: > Use an exercism/... admonition

In the past, I've failed to make external links work within admonitions. My mistake, or is this an Exercism limitation?

> Members and Extension is specific to certain IDEs, list them!

Sorry, I don't understand this. I'm talking about web pages in Chrome and Firefox.

concepts/strings/about.md
colinleach: Added to the About, this section is removed completely from the Introduction (I'm sure the exercise won't need `format()`.

concepts/strings/about.md
colinleach: I made the changes here, and also in #711. First time I heard this suggestion, but I like it!

concepts/strings/about.md
colinleach: OK, I found the [forum post](https://forum.exercism.org/t/link-references-in-special-blocks-admonitions-are-not-rendered/3803) on links in admonitions. I'll push that change a bit later.

I'm still confused about the Members an Extensions point.

concepts/strings/about.md
```diff
@@ -0,0 +1,158 @@
...
+
+## String functions
+
+Kotlin provides _many_ [`functions`][ref-string-functions] to manipulate strings.
+
+Mostly, these are [`extensions functions`][ref-extensions] rather than members of the `String` class, though this has little effect on how we use them.
+
+~~~~exercism/note
+Kotlin's rather complex [documentation][ref-string-functions] pages hide extension functions in the default view.
+Be sure to click `Members and Extensions` to expand this section.
```
SleeplessByte, line 73:
```suggestion
At moment of writing this, the most valuable content is hidden in a tab named `Members and Extensions`.
Click it to expand this section and see all the members and extensions available on the `String` class.
```

Future-proofing it.

SleeplessByte (approved):

concepts/strings/about.md
SleeplessByte: Ignore my M&E comment :D

PR #714, opened 2025-08-31
Add Nullability concept

colinleach:
This could probably be made a lot longer, but I tried to hit the main points. There are links to reference docs for students who want to explore further.

I have no idea what the exercise for this will look look, so revisions may be necessary at that point.

### DIFF
--- a/concepts/bitwise-operations/.meta/config.json
+++ b/concepts/bitwise-operations/.meta/config.json
@@ -1,0 +1,7 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [],
+  "blurb": "Kotlin provides a set of bitwise operations that allow you to manipulate the individual bits of integers."
+}
--- a/concepts/bitwise-operations/about.md
+++ b/concepts/bitwise-operations/about.md
@@ -1,0 +1,109 @@
+# About Bits
+
+Binary digits ultimately map directly to the transistors in your CPU or RAM, and whether each is "on" or "off".
+
+Low-level manipulation, informally called "bit-twiddling", is particularly important in system languages.
+
+Higher-level languages like Kotlin usually abstract away most of this detail.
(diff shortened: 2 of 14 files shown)


concepts/nullability/about.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Kotlin has a special `null` value.
+- All types _default_ to being non-nullable.
+- Most types can be _made_ nullable, but the programmer must explicitly order this.
+- Functions which might fail with an exception are likely to have a variant with a nullable return value.
+- Special, highly terse syntax tries to make testing for and responding to nulls as easy as possible.
+
+## Creating nullable variables
+
+Standard types generally have a [nullable equivalent][ref-nullable-type], with a `?` suffix: for example `String?` instead of `String`.
```
kahgoh, line 23: I read "Standard types" as the types that come with Kotlin, but I think this applies to _any_ type.

```suggestion
A type can be made [nullable][ref-nullable-type] by adding the `?` suffix (e.g. `String?` instead of `String`).
```

concepts/nullability/about.md
```diff
@@ -0,0 +1,106 @@
...
+This was easy to introduce (back in 1965), but soon led to regrets.
+The inventor later called it his ["Billion Dollar Mistake"][web-mistake].
+
+In recent decades, language designers have explored many ways to make nulls available but (mostly) safe.
+
+Kotlin has a few [general principles][ref-null-safety], which [differ significantly from Java][ref-java2kotlin].
+_The differences mean that Kotlin has special functionality to deal with Java interop, though details are beyond the scope of this Concept._
+
+- Kotlin has a special `null` value.
+- All types _default_ to being non-nullable.
```
kahgoh, line 16:
```suggestion
- All types _default_ to being non-nullable (i.e. can not be `null`) .
```

PR #715, opened 2025-08-31
Add Chars concept

colinleach:
I'm a bit confused about the way Kotlin implements Unicode as fixed-width 16-bit, so I'll ask for clarification on the [forum](https://forum.exercism.org/t/porting-java-concept-exercises-for-kotlin/17385/31) before marking this ready for review.

### DIFF
--- a/concepts/nullability/.meta/config.json
+++ b/concepts/nullability/.meta/config.json
@@ -1,0 +1,7 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [],
+  "blurb": "Kotlin provides optional nullable types, plus functionality to make nullability safe and convenient to use."
+}
--- a/concepts/nullability/about.md
+++ b/concepts/nullability/about.md
@@ -1,0 +1,106 @@
+# About Nullability
+
+In an ideal world, we could rely on everything being nicely defined, with concrete values and no gaps.
+
+This conflicts with our everyday experience, and thus many programming languages have a way to represent the absence of a value: null or NULL, nil, None, nothing...
+
+This was easy to introduce (back in 1965), but soon led to regrets.
+The inventor later called it his ["Billion Dollar Mistake"][web-mistake].
+
+In recent decades, language designers have explored many ways to make nulls available but (mostly) safe.
+
+Kotlin has a few [general principles][ref-null-safety], which [differ significantly from Java][ref-java2kotlin].
+_The differences mean that Kotlin has special functionality to deal with Java interop, though details are beyond the scope of this Concept._
+
+- Kotlin has a special `null` value.
+- All types _default_ to being non-nullable (i.e. can not be `null`) .
+- All types can be _nullable_, but this must be explicitly done.
+- Functions which might fail with an exception are likely to have a variant with a nullable return value.
+- Special, highly terse syntax tries to make testing for and responding to nulls as easy as possible.
+
+## Creating nullable variables
+
+A type can be made [nullable][ref-nullable-type] by adding the `?` suffix (e.g. `String?` instead of `String`).
+
+```kotlin
+var a = "Kotlin"  // inferred type is String
+a = null          // => Error: Null can not be a value of a non-null type String
+
+var b: String? = "Kotlin"  // nullable type
+b = null                   // => b is now null
+b == null                  // => true
+```
+
+## Useful operators and functions
+
+### The [safe call operator][ref-safe-call] `?.`
+
+For a non-nullable `String`, we can get the `.length` property.
+
+A nullable `String?` fails with the same syntax, but using `?.length` allows for the possibility of a `null`.
+The return type is `Int?`.
+
+```kotlin
+var a = "Kotlin"
+a.length  // => 6
+
+// b is still null
+b.length  // => Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type String?
+b?.length // => null
+a?.length // => 6
+```
+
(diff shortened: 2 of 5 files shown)


concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+a::class.qualifiedName  // => kotlin.Char
+a.code  // => 97
+
+val jha = 'झ'  // Devanagari alphabet
+jha.code  // => 2333
+
+val heart = '❤'  // heart emoji
+heart.code  // => 10084
+
+Char.MAX_VALUE.code  // => 65535 (64k, the largest code point allowed)
```
SleeplessByte, line 56:
```suggestion
Char.MAX_VALUE.code // => 65535 (64k, the largest code point allowed)

val not_char = 'abc' // => Too many characters in a character literal.
```

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Converting between `Char` and `Int` is simple:
```
SleeplessByte, line 59:
```suggestion
Converting between `Char` and `Int` is straightforward:
```

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+
+```kotlin
+a.code  // => 97  (a.toInt() is deprecated)
```
SleeplessByte, line 62:
```suggestion
a.code  // => 97
```

Deprecated thus will be removed at some point and then this concept becomes outdated ;)

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+
+```kotlin
+'a' + 5       // => 'f'
+'c' - 'a'   // => 2
+'c' + 'a'   // => error!
+
+'f' + ('A' - 'a')  // => 'F' (same as 'f'.uppercase()
+
+'f'.dec() // => 'e' (decrement)
+'f'.inc() // => 'g' (increment)
```
SleeplessByte, line 76:
```suggestion
'a' + 5     // => 'f'
'c' - 'a'   // => 2
'c' + 'a'   // => error!

'f' + ('A' - 'a')  // => 'F' (same as 'f'.uppercase()

'f'.dec() // => 'e' (decrement)
'f'.inc() // => 'g' (increment)
```

I am not a fan of the uppercase line. Yes, it's the same as `'f'.uppercase()`, as `'A' - 'a'` should yield `-32`, and that's ASCII, but `uppercase` is locale-aware and thus much smarter than that.

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+
+As always, there are far too [many functions][ref-char-lib] to discuss here, so this is just a selection.
+
+- For appropriate alphabets, change case with [`uppercase()`][ref-uppercase] and [`lowercase()`][ref-lowercase].
+- Test case with [`isUpperCase()`][ref-isuppercase] and [`isLowerCase()`][ref-islowercase].
+- Test character type with:
+    - [`isLetter()`][ref-isletter], covers many alphabets
+    - [`isDigit()`][ref-isdigit], in range 0..9
+    - `isLetterOrDigit()`, combines the previous two
+    - [`isWhitespace()`][ref-iswhitespace], any whitespace character
```
SleeplessByte, line 89:
```suggestion
- Test character type with:
  - [`isLetter()`][ref-isletter], covers many alphabets (the Lu, Ll, Lt, Lm, and Lo categories in unicode)
  - [`isDigit()`][ref-isdigit], in range 0..9 (the Nd category in unicode)
  - `isLetterOrDigit()`, combines the previous two
  - [`isWhitespace()`][ref-iswhitespace], any whitespace character (the Cc, Zp, Zl, and Zs categories in unicode)
```

I think this is the correct markdown format. 

For about.md only, I think we can explain how it's implemented by specifying the unicode categories (that's how these functions work!)

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+'A'.isLowerCase()   // => false
+'4'.isDigit()       // => true
+'\t'.isWhitespace() // => true  (tab character)
+```
+
+Also, [regular expressions][ref-regex] (which will be the subject of a later Concept) allow powerful search and manipulation.
+
+## Char List and String interconversions
+
+For String to Char List, we can use `toList()`.
```
SleeplessByte, line 102:
```suggestion
To convert from a `String` to a `List` of `Char`s, we can use `toList()`.
```

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+
+For Char List to String, there is the [`joinToString()`][ref-jointostring] function, which takes a separator (often the empty string) as argument.
```
SleeplessByte, line 104:
```suggestion
To convert a `List` of `Char`s to a `String`, there is the [`joinToString()`][ref-jointostring] function, which takes a separator (often the empty string) as argument.
```

concepts/chars/about.md
```diff
@@ -0,0 +1,141 @@
...
+
+```kotlin
+val kt = "kotlin".toList()  // => [k, o, t, l, i, n]
+kt.joinToString("")   // => "kotlin"
+kt.joinToString("_")  // => "k_o_t_l_i_n"
+```
+
+Note that `joinToString()` operates on a List or Array.
```
SleeplessByte, line 112:
```suggestion
Note that `joinToString()` operates on a List or Array.
```

What was the reason for this sentence?

SleeplessByte (approved):

PR #742, opened 2026-06-25
Add when concept

kahgoh:
This adds the `when` concept for Kotlin. Also took the opportunity to fix the name for the "Log Levels" exercise.

@SleeplessByte, this is ready for review


concepts/when/about.md
```diff
@@ -0,0 +1,86 @@
+# Introduction
+
+Kotlin's `when` is a conditional expression for choosing among several options.
+It is similar to `switch` in Java, but more flexible than a chain of `if` ... `else if` ... `else`.
```
SleeplessByte:
```suggestion
It is similar to `switch` in Java, and more flexible than a chain of `if` ... `else if` ... `else`.
```

concepts/when/about.md
```diff
@@ -0,0 +1,86 @@
...
+    "poodle" -> "dog"
+    else -> "unknown"
+}
+// type => "bird"
+```
+
+- Each branch is written as `pattern -> result`.
+- Branches are tried from to bottom until a match is found. Only the first branch runs.
+- Use `else` for any value not listed above (similar to the final `else` in an `if` chain).
+
```
SleeplessByte, line 27: So this is pattern matching then?

Could be something to call out!

concepts/when/about.md
```diff
@@ -0,0 +1,86 @@
...
+## Guard conditions
+
+When matching a value, additional conditions can be placed in an `if` after the pattern.
+These conditions are called guard conditions.
+
+```kotlin
+val x = 5
+val y = 10
+val description = when (x) {
+    5 if y % 2 == 0 -> "x is 5, y is even"
```
SleeplessByte, line 93: Does regexp work?

concepts/when/introduction.md
```diff
@@ -0,0 +1,86 @@
+# Introduction
+
+Kotlin's `when` is a conditional expression for choosing among several options.
+It is similar to `switch` in Java, but more flexible than a chain of `if` ... `else if` ... `else`.
```
SleeplessByte:
```suggestion
It is similar to `switch` in Java, and more flexible than a chain of `if` ... `else if` ... `else`.
```

exercises/concept/guessing-game/.docs/instructions.md
```diff
@@ -0,0 +1,58 @@
+# Instructions
+
+In this exercise, you are playing a number guessing game with a friend. 
+The rules are simple: you secretly choose a number between `1` and `100` and your friend tries to guess what number you've chosen. 
+To help your friend, you respond differently depending on how close the guess was to the number you've chosen (`42`) and how many guesses have been made. 
+These are the rules for the different replies:
+
+- If the guess is `42`: "Correct"
+- If the guess is `41` or `43` and there have been less than : "So close"
```
SleeplessByte:
```suggestion
- If the guess is `41` or `43` and there have been no more than 5 guesses: "So close"
```

exercises/concept/guessing-game/.docs/introduction.md
```diff
@@ -0,0 +1,86 @@
+# Introduction
+
+Kotlin's `when` is a conditional expression for choosing among several options.
+It is similar to `switch` in Java, but more flexible than a chain of `if` ... `else if` ... `else`.
```
SleeplessByte:
```suggestion
It is similar to `switch` in Java, and more flexible than a chain of `if` ... `else if` ... `else`.
```

SleeplessByte (approved):

concepts/when/about.md
kahgoh: Yeah, regexp can work as part of the guard. Something like this:

```kotlin
val code = 3
val classification = "animal"
val type = when (code) {
    3 if """.*mal""".toRegex().matches(classification) -> "animal"
    else -> "unknown"
}
```

But if using it match the value, I think you can only do it by matching conditions (i.e. without a value to match). For example:

```kotlin
val classification = "animal"
val type = when {
    Regex("''.*mal""").matches(classification) -> "animal"
    Regex(''"fru.*""").matches(classification) -> "fruit"
    else -> "unknown"
}
```

kahgoh: Oh, I just learn I can indeed use Regexp but would need to define an extension function:

```kotlin

operator fun Regex.contains(text: CharSequence): Boolean = this.matches(text)

fun main() {
    val classification = "animal"
    val res = when(classification) {
        in Regex(""".*mal""") -> println("animal")
        in Regex("""fru.*""") -> println("fruit")
        else -> println("unknown")
    }

    println(res)
}
```

kahgoh: Added something about this in the `about.md`.

concepts/when/about.md
kahgoh: Added mention about pattern matching.

